PR #357, opened 2020-06-30
Make API request examples match

binaryseed:
This PR makes all the API request examples on the "Collect Data" page match...

* Each example is based on `curl` using the same style
* Uses `$VARIABLES` instead of placeholders so these commands can be executed as-is w/o editing them
* Removes the python example code which is tangential to showing how to use the Events API


jerelmiller:
Thanks for submitting this @binaryseed!

binaryseed:
Thank you!

Issue #340, opened 2020-06-29
The sidenav does not stay expanded on page refresh


jerelmiller:
### Description
When navigating the site, the sidenav expands/collapses the sections as you expect when visiting pages within a section. However, if you visit a page, then refresh, you'll notice the sidebar expands every section and does not highlight the current page you are on. 

### Steps to reproduce
1. Navigate to any page except the home page
2. Notice the side nav is collapsed/expanded propertly
3. Refresh the page
4. Notice that all sections have been expanded again and the current page is not properly bolded.

### Expected behavior
Refreshing the page should not matter whether the sidenav is expanded or collapsed.

### Screenshots
**When navigating:**
<img width="1920" alt="Screen Shot 2020-06-29 at 3 15 50 PM" src="https://user-images.githubusercontent.com/565661/86061345-7c44fa00-ba1b-11ea-9cb4-7651a7a4bc42.png">

**After refreshing:**
<img width="1920" alt="Screen Shot 2020-06-29 at 3 15 57 PM" src="https://user-images.githubusercontent.com/565661/86061353-7fd88100-ba1b-11ea-929d-32c09b7a497e.png">

LizBaker:
This has been _mostly_ resolved. Though I have noticed that upon refresh you see the expanded nav blink before showing the correct collapsed/expanded state

PR #359, opened 2020-06-30
Dark mode

### LINKED ISSUE #310: Implement Dark Mode in the New Dev Ex Site
### User Story 
As a developer I would like to be able to toggle the devex site to dark mode and then back to light mode.

### Acceptance Criteria 
* Offer a Dark Mode toggle similar to the [OpenSource Site](https://github.com/newrelic/opensource-website/blob/develop/src/components/DarkModeToggle.js) 
* Offer a way to toggle light mode / dark mode in the UI 
* Toggle can go in the Right Corner of the Top Navigation

jerelmiller:
Closes #310 

## Description
Implement dark mode for the developer site. Adds the color palette and updates existing usage of colors to work with an abstracted set of values.

## Screenshot(s)
localhost_8000_ (2)


djsauble:
Default body text is too light. Try #a5bbbd instead.

Before:

Screen Shot 2020-06-30 at 2 21 11 PM

After:

Screen Shot 2020-06-30 at 2 22 25 PM

markweitzel:
The icon for dark mode needs to be consistent with the open source site.

honeyd:
To stay consistent with the open source site, should we move the dark mode toggle to the far right in the nav?

djsauble:
Looks like `pre` tags need some dark mode love (example: http://localhost:8000/components/platform-state-context)

Screen Shot 2020-06-30 at 2 28 29 PM

markweitzel:
It seems like the background "stripe" color is different b/t the os & dev sites. This may affect the color of the "featured boxes" in the middle of the page. 
New Relic Developers 2020-06-30 17-29-44

djsauble:
Given that NR1 components don't support dark mode, we shouldn't show example components on a background that changes with the dark mode toggle. Style the containing elements to always have a light background, similar to what they'd have in NR1 proper.

Screen Shot 2020-06-30 at 2 31 38 PM

markweitzel:
Confirm the primary button cta color between OS & Dev
New Relic Developers 2020-06-30 17-33-15

jerelmiller:
Another thing to note is that we are using a different code theme than the OSS site. We should unify with that theme.

djsauble:
The code snippets under the prop lists aren't contrasty enough. A shade or two lighter should do the trick.

Screen Shot 2020-06-30 at 2 37 30 PM

jerelmiller:
> Style the containing elements to always have a light background, similar to what they'd have in NR1 proper.

Do you think this should be as simple as a light gray box in light mode and the same in dark mode? We don't really have a container around the components right now, but I think that would help the situation possibly. I think the trick is making it look right in dark mode so that its apparent that the background behind them is intentional.

djsauble:
@jerelmiller Yep! The exact color doesn't matter as much as it being consistent between modes (to emphasize the dark-on-light look these components will have in-product)

djsauble:
Is it possible to theme the icons? They have a great hover state in light mode, but it all kind of falls apart in dark mode.

Screen Shot 2020-06-30 at 2 48 06 PM

jerelmiller:
Ooops, the icon gallery was an oversight. Good catch!

djsauble:
The prop list deprecation notices don't have dark mode styling. Suggest light text on a dark red background.

Screen Shot 2020-06-30 at 2 52 22 PM

djsauble:
The callouts also lack a light-on-dark dark mode theme.

Screen Shot 2020-06-30 at 2 55 26 PM

djsauble:
Hamburger needs dark mode styling

Screen Shot 2020-07-01 at 9 30 08 AM

jerelmiller:
@djsauble Fixed with 6ea0fab

src/components/GlobalHeader.js
```diff
@@ -66,15 +68,23 @@ const GlobalHeader = ({ className }) => {
...
           )}
           <li className={styles.rightSideButton}>
             <ExternalLink href={`${githubBaseUrl}/issues/new/choose`}>
-              <FeatherIcon name="github" size="1rem" />
+              <FeatherIcon name="github" size="1rem" className={styles.icon} />
             </ExternalLink>
           </li>
+          <li className={styles.rightSideButton}>
+            <FeatherIcon
+              name={darkMode.value ? 'sun' : 'moon'}
```
zstix, line 82: Nitpick: I noticed that, when the site first loads, the moon icon is displayed momentarily. I suspect this is due to `darkMode.value` being falsy. We may want to not show the icon at all until we have access to that variable. It seems like the OSS site operates this way.

jerelmiller:
You're right. Good catch. I need to do a bit of research to determine how to avoid this scenario. I suspect its because this page is "server rendered", with no access to local storage or media queries.

jerelmiller:
> I also noticed that the sidebar then expands all the way when refreshing

This bug is logged in #340 and outside the scope of this PR since its an existing bug.

src/components/GlobalHeader.js
jerelmiller: I might log this as a separate issue. I think it has to do with the fact that its a pre-rendered page, which doesn't have access to the local storage/media queries. Even if we hide it, there will still be a flash of content, which would be nice to avoid altogether.

Thoughts? Or should I try and solve this in this PR?

jerelmiller: Turns out it's more complicated than I thought. I logged this issue in #385. Its annoying, but not a blocker for now, so lets fix at a later time. This will allow us to get this out the door sooner.

PR #384, opened 2020-07-01
Updated how the sidebar is positioned

### LINKED ISSUE #327: [Stretch] On iOS, the global navigation can be "pulled" over the site content
### Description 
When "pulling" down on a page to trigger a refresh, you can pull the global navigation (top-bar) off the top of the page and over the main content area. See attached screenshot. 
image2

### Steps to Reproduce: 
1. Go to the site on a iOS device 
2. Pull down on the page as if to refresh it

### Expected Result
The top-bar should be locked to the top. When pulling down on the page to refresh, the whole page should be pulled down.

zstix:
## Description
On iOS, you can "pull down" on the page to refresh the site. When you do this, the site and top-bar navigation get pulled down, but the sidebar does _not_. This is because the sidebar had `position: fixed;`. I've updated it to use an abolute positioning-based approach. There are some caveats with this approach (see next section).

## Reviewer Notes
By removing `position: fixed;` we are slightly changing how the sidebar scrolls. This is a draft PR because I would like some :eyes: on this to ensure we are alright with the new scrolling behavior.

Fixes #327

### DIFF
--- a/src/components/Layout.module.scss
+++ b/src/components/Layout.module.scss
@@ -11,6 +11,7 @@
   display: grid;
   grid-template-columns: var(--sidebar-width) minmax(0, 1fr);
   width: 100%;
+  position: relative;
 
   @media screen and (max-width: 760px) {
     grid-template-columns: minmax(0, 1fr);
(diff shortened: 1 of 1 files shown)


zstix:
Checking to see if we want to go this route. We may close this PR and mark the issue as `Wont fix`

zstix:
We are going to punt on this.

PR #386, opened 2020-07-01
Align images to the right in steps

### LINKED ISSUE #312: Can't move image to the right column of a <Step>
### Description 

There’s no apparent way to move an image to the right column in blocks.

### Steps to Reproduce 
* Create a block
* Add an image anywhere 
* Check layout

### Expected behavior 
I would expect that images appear on the right side, or that at least can be moved there if needed, with automatic resizing (kind of what happens with videos in the Intro block).

### Screenshots
image1

LizBaker:
## Description
This PR adds images to the list of elements to right align in steps and deletes the guide example.

Gatsby adds a wrapper to images as well as an anchor tag, for simplification I have removed this anchor tag in order to make the search for images more shallow

Closes #312 

## Screenshot(s)
Screen Shot 2020-07-01 at 1 11 43 PM


src/components/SideBySide.js
```diff
@@ -9,13 +9,27 @@ import { isMdxType } from '../utils/mdx';
 const SideBySide = ({ className, children, type }) => {
   const types = Array.isArray(type) ? type : [type];
   const childObjects = Children.toArray(children);
+
+  const isGatsbyMdxType = (child, type) => {
+    if (type === 'img' && isMdxType(child, 'span')) {
+      if (Array.isArray(child.props.children)) {
```
zstix, line 15: Nitpick: We can probably combine these two `if` statements with `&&`.

src/components/SideBySide.js
```diff
@@ -9,13 +9,27 @@ import { isMdxType } from '../utils/mdx';
...
+        return (
+          child.props.children.filter((child) => isMdxType(child, 'img'))
+            .length > 0
```
jerelmiller:
```suggestion
          child.props.children.some((child) => isMdxType(child, 'img'))
```

Rather than filtering and then checking length, you can use `.some` which will return `true` if at least one of the items is `true`

src/components/SideBySide.js
```diff
@@ -9,13 +9,29 @@ import { isMdxType } from '../utils/mdx';
 const SideBySide = ({ className, children, type }) => {
   const types = Array.isArray(type) ? type : [type];
   const childObjects = Children.toArray(children);
+
+  const isGatsbyMdxType = (child, type) => {
```
jerelmiller, line 13: This name is a bit misleading and doesn't really tell me how it differentiates with `isMdxType`. Is there something we can do to make this more obvious at what this function does?

[nit] Since this function does not use any of its outer scope, you can pull it out of the component.

src/components/SideBySide.js
LizBaker: Oh nice, that's handy

src/components/SideBySide.js
LizBaker: Hmm yeah, I named it this way because it's still supposed to ultimately be checking if the type we're passing in matches the type of the child we're looking at. But since we know the images will always be nested we have to add this extra check. Happy to rename it, but having a hard time! any suggestions?

PR #388, opened 2020-07-01
fix: Reduce spacing between guide step and title

djsauble:
## Description

The guide step number and title should be connected visually, which we accomplish by reducing the spacing between these elements.

### Before

Screen Shot 2020-07-01 at 2 55 02 PM

### After

Screen Shot 2020-07-01 at 2 55 10 PM


jerelmiller:
Thanks for updating this! That does look much better.

PR #399, opened 2020-07-02
feat: Add cookies dialog

### LINKED ISSUE #303: Provide cookie consent confirmation
### Description 

The OSS put intercom in and it will be released with that widget, but Joel has taken it out  

We don't seem to be using it the Core Docs site, but it's on the old Dev Site and the marketing site. 

So in the short term to get a cookie consent message into the new DevSite we have decide to leverage what Daniel Golden is building and hopefully can borrow that code (see attached video)

### User Story 

As a user I should be clearly told the DevSite uses Cookies and have option of accepting or denying this.

### Implementation Details: 
* Use the same experience found on the OSS site. (opensource.newrelic.com)
* Use the same gatsby plugin that the OSS site uses: https://www.gatsbyjs.org/packages/gatsby-plugin-gdpr-cookies/
* Prompt a user when they first go to the site and let them know about cookies.
* If they accept, allow cookies to be used
* if they reject, disable cookies from being used.

* See the OSS Repo for more implementation details. 

https://github.com/newrelic/opensource-website/pull/484
https://github.com/newrelic/opensource-website/pull/522/files
https://github.com/newrelic/opensource-website/tree/danielgolden-feat/cookie-approval-dialog

### Release coordination 

We will want to release this feature with the OSS Site team on the same day when we return from the long 4th O' July weekend. Ideally we'd release on July 6th / 7th

timglaser:
Closes #303 

Ready for review, but **DO NOT MERGE YET** as this is to be released at the same time as the same dialog on the open source site. That is why this PR is marked as blocked.

## Description
Displays a cookie acceptance dialog if you have not provided an answer in a prior visit. Handles light and dark mode.

## Reviewer Notes
You can play around with this by looking at the Application tab in Chrome's dev tools and clearing out the cookie (`newrelic-gdpr-consent`) or changing its value.

## Screenshot(s)
Screen Shot 2020-07-09 at 3 19 44 PM


gatsby-browser.js
```diff
@@ -6,4 +6,24 @@
...
+const onRouteUpdate = ({ location }) => {
+  if (process.env.NODE_ENV !== `production` || typeof ga !== `function`) {
+    return null;
+  }
+
+  // wrap inside a timeout to make sure react-helmet is done with it's changes (https://github.com/gatsbyjs/gatsby/issues/9139)
+  // reactHelmet is using requestAnimationFrame: https://github.com/nfl/react-helmet/blob/5.2.0/src/HelmetUtils.js#L296-L299
+  const sendPageView = () => {
+    const pagePath = location
+      ? location.pathname + location.search + location.hash
```
zstix, line 18: Will we always have these properties on the `location` object? Just wanna make sure we're not adding `undefined` to the path.

gatsby-browser.js
```diff
@@ -6,4 +6,24 @@
...
+      : undefined;
+    window.ga(`set`, `page`, pagePath);
+    window.ga(`send`, `pageview`);
+  };
+
+  // Minimum delay for reactHelmet's requestAnimationFrame
```
zstix, line 24: Would you be willing to explain this a little more. I'm unclear about why this necessary.

src/components/CookieApprovalDialog.js
```diff
@@ -0,0 +1,82 @@
+import React, { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+import Cookies from 'js-cookie';
+import Button from './Button';
+import styles from './CookieApprovalDialog.module.scss';
```
zstix, line 5: We _could_ move to emotion for the styles. We could also totally wait.

src/components/CookieApprovalDialog.js
```diff
@@ -0,0 +1,82 @@
...
+      options.domain = 'newrelic.com';
+    }
+
+    Cookies.set(gdprConsentCookieName, String(!!answer), options);
+    setIsCookieSet(true);
+    answer && setCookieConsent(true);
+  }
+
+  return (
+    !isCookieSet && (
```
zstix, line 30: We may want to just do something like `if (isCookieSet) return null;` above this return statement. That way, we cut down on one more indent level.

src/components/CookieApprovalDialog.js
```diff
@@ -0,0 +1,82 @@
...
+      <div className={`${styles.container} ${className || ''}`}>
```
zstix, line 31: If we're _not_ using emotion, we could still clean this up with classnames:
```<div className={cx(styles.container, className)}>```

File: src/components/CookieApprovalDialog.js [hunk 6]
```diff
@@ -0,0 +1,82 @@
...
+              <span role="img" aria-label="Cookie emoji">
+                🍪
+              </span>
+            </h4>
+            <p className={styles.description}>
+              We use cookies and other similar technologies ("Cookies") on our
+              websites and services to enhance your experience and to provide
+              you with relevant content. By using our websites and services you
+              are agreeing to the use of cookies. You can read more{' '}
+              <a
```
zstix, line 45: We have an `ExternalLink` component that we might want to consider using here.

gatsby-browser.js
LizBaker: apparently it's a common issue that google analytics will fire the pageview off before helmet has finished injecting the page titles, so folks were seeing improperly tracked pageviews. this is gatsby's current answer for dealing with that

gatsby-browser.js
LizBaker: I believe so yes, this is used in gatsby's google analytics plugin

src/components/CookieApprovalDialog.js
LizBaker: I want to wait on that for this PR, but soon!

PR #400, opened 2020-07-02
Added Tessen to Fix Google Analytics

### LINKED ISSUE #394: Google Analytics isn't reporting data for the new Dev Site
### Description

After launching the new site, it doesn't look like GA is reporting on site traffic.

### Steps to reproduce

1. Go to or instance of GA: (if you need to view GA,  JV can provide the dashboard link)
2. Look at the reports for site traffic
3. See no data being reported

### Expected behavior

Site traffic is reported correctly in GA.

### Screenshots

Screen Shot 2020-07-01 at 5 43 24 PM

Screen Shot 2020-07-02 at 10 10 54 AM

zstix:
## Description
Google Analytics is under-reporting pageviews. It turns out this is because we were using _Tessen_ to track pageviews and send that information to GA via Google Tag Manager. This PR _should_ resolve this.

## Reviewer Notes
I'm not 100% sure I know what I am doing here. Google Analytics will need 24 hours before we can tell if it's recording pageviews correctly. If anyone has any suggestions for how to get this working, I'm all :ear:s.

Closes #394

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "private": true,
   "version": "0.1.0",
   "dependencies": {
+    "@datanerd/tessen": "^1.3.0",
     "@mdx-js/mdx": "^1.6.4",
     "@mdx-js/react": "^1.6.5",
     "classnames": "^2.2.6",
--- a/gatsby/wrap-page-element.js
+++ b/gatsby/wrap-page-element.js
@@ -5,9 +5,11 @@ import { PageContext } from '../src/components/PageContext';
 
 import createBreadcrumbs from '../src/utils/create-breadcrumbs';
 import pages from '../src/data/sidenav.json';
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -3,6 +3,7 @@
   "private": true,
   "version": "0.1.0",
   "dependencies": {
+    "@datanerd/tessen": "^1.3.0",
```
jerelmiller, line 6: This is going to cause issues for anybody not on the New Relic VPN (all non-employee contributors) since `@datanerd` does not live in NPM. How do we best approach this so that we can still have external contributors to the project?

zstix:
Apparently we should _not_ be using Tessen and just use the NR Browser agent for this. Not sure how to get that working. Either way, this PR is no long applicable.

PR #411, opened 2020-07-08
GitHub Action to Generate Changelog

### LINKED ISSUE #301: [Open source checklist] Add auto generated CHANGE_LOG.md
### Description 
This ticket is created in reaction to Open Source Release Checklist.

Inspiration could be drawn from the actions based approach in the open source site's repo. You can see the change log history commits come from the bots https://github.com/newrelic/opensource-website/commits/develop/CHANGELOG.md

### Acceptance Criteria

* Merged to master triggers an automatic commit to update the change log based on our semantic commits.
* Only commits of `fix` or `feat` type show up in the change log.

zstix:
## Description
Adds a GitHub Action to generate a [semantic release](https://github.com/semantic-release/semantic-release) (and generate a changelog) whenever we merge into `master`. Much of this was taken from [what the OSS had set up](https://github.com/newrelic/opensource-website/blob/develop/.github/workflows/release.yml#L114-L153).

## Reviewer Notes
This will get kicked off when it gets merged into master. The only thing I was unsure about was the `Cache node_modules` step.

Closes #301

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,52 @@
+name: Changelog
+
+on:
+  push:
+    branches:
+      # todo: change this branch name
+      - master
+    repository_dispatch:
+      types: [semantic-release]
+
+env:
+  THIRD_PARTY_GIT_AUTHOR_EMAIL: opensource+bot@newrelic.com
+  THIRD_PARTY_GIT_AUTHOR_NAME: nr-opensource-bot
+
+jobs:
+  generate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          # semantic-release requiers >= 10.18
+          node-version: 12
+
+      # todo: do we need this?
+      - name: Cache node_modules
+        id: cache-node-modules
+        uses: actions/cache@v2
+        env:
+          cache-name: node-modules
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ env.cache-name }}-
+
+      # todo: update / add this
+      - name: Install Dependencies
+        run: npm ci
+
+      - name: Semantic Release
+        env:
+          # Use nr-opensource-bot for authoring commits done by
+          # semantic-release (rather than using @semantic-release-bot)
+          GIT_AUTHOR_NAME: "nr-opensource-bot"
+          GIT_AUTHOR_EMAIL: "opensource+bot@newrelic.com"
+          GIT_COMMITTER_NAME: "nr-opensource-bot"
+          GIT_COMMITTER_EMAIL: "opensource+bot@newrelic.com"
+        run: npx semantic-release
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -5,8 +5,6 @@ on:
     branches:
       # todo: change this branch name
       - master
-    repository_dispatch:
-      types: [semantic-release]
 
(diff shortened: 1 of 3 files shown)


.github/workflows/changelog.yml
```diff
@@ -0,0 +1,51 @@
...
+  push:
+    branches:
+      # todo: change this branch name
+      - master
+    repository_dispatch:
+      types: [semantic-release]
+
+env:
+  THIRD_PARTY_GIT_AUTHOR_EMAIL: opensource+bot@newrelic.com
+  THIRD_PARTY_GIT_AUTHOR_NAME: nr-opensource-bot
```
jerelmiller, line 13: It doesn't look like you're using these environment variables in this action, so feel free to delete these.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,51 @@
+name: Changelog
+
+on:
+  push:
+    branches:
+      # todo: change this branch name
+      - master
+    repository_dispatch:
+      types: [semantic-release]
```
jerelmiller, line 9: Can you talk about what this does? I'm not sure I fully understand how `repository_dispatch` works.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,51 @@
+name: Changelog
```
jerelmiller, line 1: Out of curiosity, why did you choose to make this a changelog workflow rather than a release workflow (by naming that is)?

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,51 @@
...
+  generate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          # semantic-release requiers >= 10.18
```
jerelmiller, line 25: [typo] `requires`

.github/workflows/changelog.yml
zstix: I was under the impression that `name` was simply what was displayed in the GitHub UI and I thought that a `Changelog` was more descriptive. If there are other implications for this field, I can adjust this.

.github/workflows/changelog.yml
zstix: I brought this over from what the OSS was doing. I believe it's required to allow actions to have the right scope for writing to the repository, but I'm not clear. [Here](https://github.com/marketplace/actions/repository-dispatch) is the action details.

.github/workflows/changelog.yml
jerelmiller: Oops sorry I should have been more specific, I was really commenting on the file as a whole (I wish there was a better way to do that in GitHub). I wasn't talking necessarily about this particular field, but rather this workflow in general. It looks like the open source site has this under their `release` workflow alongside some of the other release related items like generating the third party notices. 

I was mostly just curious why this was its own workflow instead of part of a release workflow.

.releaserc
```diff
@@ -0,0 +1,21 @@
+{
+  "branches": ["develop"],
```
jerelmiller, line 2: This should be `master` for now since we don't yet have a `develop` branch.

.github/workflows/release.yml
```diff
@@ -0,0 +1,45 @@
...
+          path: ~/.npm
+          key: ${{ runner.os }}-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ env.cache-name }}-
+
+      # deletes node_modules, install dependencies, does NOT update lockfile
+      - name: Install Dependencies
+        run: npm ci
+
+      - name: Semantic Release
```
jerelmiller, line 37: Does this step need the bot token?

zstix: It does and I have no idea how I missed that! 😵

PR #418, opened 2020-07-09
Setup emotion and convert Button

jerelmiller:
## Description
Configures emotion for use within the site and jest. Also converts the `<Button />` component to use emotion rather than CSS modules.


caylahamann:
Would we be able to do a walk through of this just so I know some best practices for using emotion?

jerelmiller:
Totally!

PR #427, opened 2020-07-10
Image component for markdown components 

### LINKED ISSUE #328: [stretch] Make a component to standardize image sizes for guides
### Description 
I think we should standardize on the image size and then build a component for that eventually.

caylahamann:
## Description
This PR creates an image component for markdown components. It provides the same functionality as `![]` syntax, but the hope is that we can expand on this to standardize our images"

## Reviewer Notes
I replaced one component in one guide to demonstrate the difference. If you look at "Create a 'Hello, World!' application guide underneath the Build Apps section. 

Closes #328

### DIFF
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -1,8 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-const Image = ({ src, alt, width }) => {
-  return <img src={src} alt={alt} width={width} />;
+const Image = (props) => {
+  return <img {...props} />;
 };
 
 Image.propTypes = {
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -1,32 +1,18 @@
 import React from 'react';
-import { useStaticQuery, graphql } from 'gatsby';
-import Img from 'gatsby-image';
+import PropTypes from 'prop-types';
 
-/*
- * This component is built using `gatsby-image` to automatically serve optimized
- * images with lazy loading and reduced file sizes. The image is loaded using a
- * `useStaticQuery`, which allows us to load the image from directly within this
- * component, rather than having to pass the image data down from pages.
- *
- * For more information, see the docs:
- * - `gatsby-image`: https://gatsby.dev/gatsby-image
- * - `useStaticQuery`: https://www.gatsbyjs.org/docs/use-static-query/
- */
+const Image = ({ src, alt, width }) => {
+  return <img src={src} alt={alt} width={width} />;
+};
 
-const Image = () => {
-  const data = useStaticQuery(graphql`
-    query {
-      placeholderImage: file(relativePath: { eq: "gatsby-astronaut.png" }) {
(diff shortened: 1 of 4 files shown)


src/components/Image.js
```diff
@@ -1,32 +1,18 @@
...
- * images with lazy loading and reduced file sizes. The image is loaded using a
- * `useStaticQuery`, which allows us to load the image from directly within this
- * component, rather than having to pass the image data down from pages.
- *
- * For more information, see the docs:
- * - `gatsby-image`: https://gatsby.dev/gatsby-image
- * - `useStaticQuery`: https://www.gatsbyjs.org/docs/use-static-query/
- */
+const Image = ({ src, alt, width }) => {
+  return <img src={src} alt={alt} width={width} />;
```
zstix, line 5:
```suggestion
const Image = (props) => {
  return <img {...props} />;
```

caylahamann: when I try to do this, i get a linting error because the linter wants there to be an alt prop, so I think i'm just going to keep it in the structure that I currently have it in

jerelmiller: @caylahamann You can get around the lint error and apply @zstix's suggestion by destructuring the `alt` prop and spreading the rest of them like this:

```js
const Image ({ alt, ...props }) => (
  <img alt={alt} {...props} />
)
```

src/components/Image.js
```diff
@@ -1,32 +1,18 @@
...
-  `);
+Image.propTypes = {
+  src: PropTypes.string.isRequired,
+  alt: PropTypes.string.isRequired,
+  width: PropTypes.number,
+};
 
-  return <Img fluid={data.placeholderImage.childImageSharp.fluid} />;
+Image.defaultProps = {
+  width: 1200,
```
jerelmiller, line 15: Would it make more sense to make this width `auto` by default? Seems weird we'd want to specify a hard width for all images rather than let theme be their natural size until someone configures it otherwise. Thoughts?

caylahamann: So my understanding of the ticket was that we wanted to standardize the size of the images (?) @mmfred @djsauble what are your thoughts on this?

jerelmiller: Great point 😄 . I'll be curious to see what the others say .

djsauble: Images that are too large in the vertical dimension make it hard to follow the flow of the page. For example:

Screen Shot 2020-07-13 at 10 52 14 AM

In contrast, images that take the full width of the page can be fine as long as they're not too tall. For example:

Screen Shot 2020-07-13 at 10 56 14 AM

My suggest is to 1) set a max-height of 400px for images, 2) let the width be up to 100% while respecting the aspect ratio of the image, 3) left align images so they follow the shape of the text.

caylahamann: Okay! thank you for those suggestions @djsauble

PR #431, opened 2020-07-11
added a link to nerdpack layout repos

jpvajda:
## Description
added a link to nerdpack layout repos.

## Reviewer Notes
n/a

n/a

## Screenshot(s)
n/a

### DIFF
--- a/src/components/__tests__/__snapshots__/Footer.js.snap
+++ b/src/components/__tests__/__snapshots__/Footer.js.snap
@@ -1,0 +1,47 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Footer renders correctly 1`] = `
+<footer
+  className="footer"
+>
+  <div
+    className="container container"
+  >
+    <div
+      className="leftColumn"
+    >
+      <a
+        className="logo logo"
+        href="/"
+      />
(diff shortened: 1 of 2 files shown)


src/components/__tests__/__snapshots__/Footer.js.snap
```diff
@@ -0,0 +1,47 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
caylahamann, line 1: I don't really know that much about Jest (doing some documentation reading right now!) but why is this included in this PR?

jerelmiller: @caylahamann I can take a guess why this is here. As part of the [emotion PR](https://github.com/newrelic/developer-website/pull/418), I [removed the line](https://github.com/newrelic/developer-website/pull/418/files#diff-a084b794bc0759e7a6b77810e01874f2L72) that gitignored the snapshot tests because, as per Jest's documentation, [these should be committed](https://jestjs.io/docs/en/snapshot-testing#should-snapshot-files-be-committed). It looks like maybe John had tests running, which committed the snapshot.

I'm ok having this here since this should have been here all along.

Issue #456, opened 2020-07-17
Navigation breadcrumb bolding doesn't maintain state after a browser refresh


jpvajda:
### Description

If you navigate down the navigation tree, each sub nav element will bold to indicate to a user where they are in the navigation. However on refresh, we don't show the bolding on all the navigation elements. 

### Steps to reproduce

1. Go to https://developer.newrelic.com/
2. Navigate to:  component library > Feedback > Spinner
3. Click refresh 
4. See how the bolding is lost on the component library nav element.

1. Go to https://developer.newrelic.com/
2. Go to Build Apps > Create a "Hello, World app" 
3. Click refresh 
4. See how the bolding is lost on the Build apps and Create "Hello, World app" nav element.

### Expected behavior

Bolding of navigate state breadcrumbing is maintained on refresh of the browser. Note this is also an issue if a user was to share a URL with another user and they opened the URL, the breadcrumbing state would not be maintained. 

### Screen shots 

<img width="321" alt="Screen Shot 2020-07-16 at 6 59 52 PM" src="https://user-images.githubusercontent.com/1395158/87738862-94e33d00-c79b-11ea-99b9-f3ba84831fdb.png">

LizBaker:
One way of getting Component Library to bold could be fixed by changing the URL path for it and all the components. The partiallyActive prop on gatsby Link will be true (and can be styled) if the current page's path matches part of the `to` prop. if the component library overview lived at `/explore-docs/components` and the components were `/explore-docs/components/spinner' then component library would get picked up as partially active when you're on a component page. not sure this is worth it, that would be a lot of redirects to worry about

jerelmiller:
We've changed the styling and are no longer bolding the text. Going to go ahead and close this.

PR #457, opened 2020-07-17
Persist bold style in Link nav items

LizBaker:
## Description
While we still don't have a full fix for #456, this PR at least maintains the bold styling on non toggle-able nav items (collect data, automate workflows, build apps) after refresh

## Reviewer Notes
Try navigating to a page within collect data, automate workflows, or build apps. Refresh the page to see the bold styling preserved

Partial fix for #456

### DIFF
--- a/src/components/NavigationItems.module.scss
+++ b/src/components/NavigationItems.module.scss
@@ -80,7 +80,6 @@ button.navLink {
   }
 }
 
-.isCurrentPage,
 .isBreadCrumb {
   font-weight: bold !important;
 }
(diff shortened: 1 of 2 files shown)


src/components/NavigationItems.module.scss
```diff
@@ -80,7 +80,6 @@ button.navLink {
   }
 }
 
-.isCurrentPage,
```
jerelmiller, line 83: Is there another definition for this class? Or does removing this mean there is no longer a need for this class. I see its still being used in the `.js` file so I'm wondering if we should remove that as well.

LizBaker: there is yes, it's used to check that anything that is *not* isCurrentPage gets a hover style

PR #458, opened 2020-07-21
Podcasts page

thejonanshow:
## Description
We're getting into the podcasting business as of tomorrow, so we need a place for those to live. I'm not sure this is the
right place in the navigation considering it sits in a collection of external links (maybe this needs to be 'podcasts.newrelic.com' eventually), but it seems reasonable enough for now. I'm happy to move it if someone feels strongly about it, but the episode pointing back here drops tomorrow so we should put it up somewhere on developer.newrelic.com promptly.

Screen Shot 2020-07-20 at 7 33 58 PM


jerelmiller:
Hey @thejonanshow! Thanks so much for your contribution! We'd love to figure out how to get this in!

@djsauble Would you mind taking a look at this PR from an IA perspective?  
@mmfred Would you mind looking at this PR from a content perspective?

src/pages/podcasts.module.scss
```diff
@@ -0,0 +1,43 @@
...
+  text-align: center;
+}
+
+.player {
+  overflow: hidden;
+}
+
+.player iframe {
+  width: 100%;
+  height: 421px;
```
jerelmiller, line 33: It looks like this height is adding a bunch of whitespace between the player and the footer. 

Screen Shot 2020-07-20 at 9 58 32 PM

It looks like the visible part of the player is `200px` tall, but also looks like there is some kind of invisible content that causes the content to scroll if this is set to `200px`. Do you know if there is a way to configure the player to reduce the height of the player to the visible portion? If possible, it would be great to reduce some of the whitespace here.

src/pages/podcasts.module.scss
```diff
@@ -0,0 +1,43 @@
...
+  display: grid;
+  grid-template-columns: repeat(2, calc(50% - 1rem));
+  grid-gap: 2rem;
+}
+
+.img {
+  width: 100%;
+}
+
+.podcasts {
```
jerelmiller, line 17: It looks like this class is not currently used. Could you remove this?

src/pages/podcasts.js
```diff
@@ -0,0 +1,67 @@
...
+          </p>
+        </div>
+        <img
+          className={styles.img}
+          src={podcastsHeader}
+          alt="podcasts header"
+        />
+      </section>
+      <section className={cx(styles.section, styles.player)}>
+        <div id="buzzsprout-player">
```
jerelmiller, line 56: Out of curiosity, is this `id` needed on this `<div />`? Could you get away with embedding the `iframe` directly inside the `section`?

src/pages/podcasts.module.scss
```diff
@@ -0,0 +1,43 @@
...
+  display: grid;
+  grid-template-columns: repeat(3, 1fr);
+  grid-gap: 2rem;
+}
+
+.point {
```
jerelmiller, line 23: It looks like this class is not currently used. Could you remove this?

src/components/GlobalHeader.js
```diff
@@ -59,6 +59,14 @@ const GlobalHeader = ({ className }) => {
                 Community
               </ExternalLink>
             </li>
+            <li>
+              <ExternalLink
+                href="https://developer.newrelic.com/podcasts"
```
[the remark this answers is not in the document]
thejonanshow: Done, thanks for including the instructions.

src/pages/podcasts.module.scss
```diff
@@ -0,0 +1,43 @@
...
+}
+
+.player iframe {
+  width: 100%;
+  height: 421px;
+  border: none;
+}
+
+
+.pointIcon {
```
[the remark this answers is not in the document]
thejonanshow: Is it obvious I copied this from the champions page? Removed.

src/pages/podcasts.js
thejonanshow: It's passed through in the URL to buzzsprout so I'm assuming it's useful for their styles or js, even though it's not obvious to me how they're using it. I'd rather leave it for now but if you feel strongly about it let me know and I'll take it out.

src/pages/podcasts.module.scss
thejonanshow: The iframe has an episode list that we're not seeing as we only have the one episode. For now I've hidden the windows related to the list so I'm able to reduce the height and get rid of the whitespace. It was bothering me as well.

thejonanshow:
Thank you for the prompt review @jerelmiller, I really appreciate it.

src/pages/podcasts.js
jerelmiller: Totally fine to leave it! If it seems like it has a purpose, best not to try and break it 😄

src/components/GlobalHeader.js
```diff
@@ -59,6 +59,11 @@ const GlobalHeader = ({ className }) => {
                 Community
               </ExternalLink>
             </li>
+            <li>
+              <Link href="/podcasts" className={styles.leftSideLink}>
```
jerelmiller, line 63:
```suggestion
              <Link to="/podcasts" className={styles.leftSideLink}>
```

It looks like the `href` prop works, but it's still causing a page refresh. The `to` prop should make it work with the SPA.

djsauble:
@thejonanshow The items in the top nav represent distinct sites. If podcasts are going to have their own site, this makes sense, but your screenshot makes me think they're intended to be part of the developer site, in which case I'd suggest adding podcasts to the sidebar nav instead.

We have a section for Developer Champions on the landing page, it'd be awesome to have a similar section for Podcasts as well to raise visibility. Something like the following?

Group

If you go this route, I'd suggest putting the postcast sidebar link immediately after the developer champions link for consistency.

jerelmiller:
@jpvajda Can you check this out as well?

jpvajda:
Before we merge, I'm going to confirm with @markweitzel  and @veextee  if this is something we want to add into the Developer site in this particular way. Thanks for submitting @thejonanshow.

thejonanshow:
> @thejonanshow The items in the top nav represent distinct sites. If podcasts are going to have their own site, this makes sense, but your screenshot makes me think they're intended to be part of the developer site, in which case I'd suggest adding podcasts to the sidebar nav instead.
> 
> We have a section for Developer Champions on the landing page, it'd be awesome to have a similar section for Podcasts as well to raise visibility. Something like the following?
> 
> Group
> 
> If you go this route, I'd suggest putting the postcasts sidebar link immediately after the developer champions link for consistency.

I like this idea for now, when we have more we could move to a separate site if it seems necessary.

We posted on the blog about the podcast about 20mins ago, maybe we could ship this and I'll put up another PR to move it next to the champions?

thejonanshow:
> > @thejonanshow The items in the top nav represent distinct sites. If podcasts are going to have their own site, this makes sense, but your screenshot makes me think they're intended to be part of the developer site, in which case I'd suggest adding podcasts to the sidebar nav instead.
> > We have a section for Developer Champions on the landing page, it'd be awesome to have a similar section for Podcasts as well to raise visibility. Something like the following?
> > Group
> > If you go this route, I'd suggest putting the postcasts sidebar link immediately after the developer champions link for consistency.
> 
> I like this idea for now, when we have more we could move to a separate site if it seems necessary.
> 
> We posted on the blog about the podcast about 20mins ago, maybe we could ship this and I'll put up another PR to move it next to the champions?

Ignore me, I'll rework it and try again.

thejonanshow:
I ended up stacking the tiles vertically as the text looked a bit crowded when they were side-by-side. I quite like the look of it but let me know if you'd prefer I go with the recommendation above.

I've also removed the CSS we were using to hide the episode list in the player as we're preparing to drop our second episode tomorrow, so something should be appearing in that extra whitespace shortly.

Screen Shot 2020-07-28 at 1 01 40 PM

djsauble:
@thejonanshow The stacked tiles are a good look, thanks for the update!

jerelmiller:
Woot! This looks awesome! Thanks so much for updating this!

PR #491, opened 2020-07-27
Adds Tessen Library

### LINKED ISSUE #472: [Setup] Add Tessen to the developer site
## Description
We would like to use _Tessen_ to help us track page views and to provide a system for us to track whether search results / related content is helpful to the user or not. We will want to download the library, add the version to the filename, and add it to the repository. We can then load it in the `<head>`.

We had [previously looked into adding Tessen](https://github.com/newrelic/developer-website/issues/394) and ran into some hurdles:
* Not having access to internal npm registries
* Not having an easy way to add the bundle to the page
* The page loading slower with the extra bundle

## Helpful Links
* #394 (previous attempt to add Tessen for GTM)

## Screenshots
N/A

zstix:
## Description
Adds _Tessen_ to the project so that we can start sending additional analytics to New Relic.

## Reviewer Notes
There are were couple of hurdles to getting _Tessen_ added to the site:
* Due to this project being open source, we could not rely on internal registries for `npm install`
* The library does not seem to play well with webpack, so `import` was not working
* The library relies on `window` and `document`, but I was unable to get it to work in any of the `gatsby-` files

It turns out that gatsby provides a [static folder](https://www.gatsbyjs.org/docs/static-folder/) for these edge cases that can be referenced using [`withPrefix`](https://www.gatsbyjs.org/docs/gatsby-link/#add-the-path-prefix-to-paths-using-withprefix). This adds it to the page header and doesn't throw any errors.

* Closes #472 (issue to add Tessen)
* #394 (previous issue to add Tessen for GTM)

## Screenshot(s)
Screen Shot 2020-07-27 at 3 58 11 PM


jerelmiller:
Out of curiosity, does this library set a global variable on `window`? If so, should we consider adding an [`externals` mapping to webpack](https://webpack.js.org/configuration/externals/) to allow us to "import" tessen? This might be nicer and feel more consistent than trying to use `window.Tessen` everywhere.

gatsby-node.js
```diff
@@ -84,3 +84,11 @@ exports.onCreateNode = ({ node, actions }) => {
     });
   }
 };
+
+exports.onCreateWebpackConfig = ({ actions }) => {
+  actions.setWebpackConfig({
+    externals: {
+      Tessen: 'Tessen',
```
jerelmiller, line 91:
```suggestion
      tessen: 'Tessen',
```

PR #496, opened 2020-07-28
Use `<CodeBlock>` component from theme

### LINKED ISSUE #494: Text selection color for Code Block is invisible (light mode only)
### Description

When browsing the component docs on the developer site, you can select text within the Code Block component. Unfortunately, the text selection / "highlight" color is not visible while on the light theme. This is _not_ an issue on the dark theme.

### Steps to reproduce

1. Go to a component doc page.
2. Ensure you are on the light mode.
3. Select text within Code Block.

### Expected behavior

This color should be visible so the user knows what they are currently selecting.

zstix:
## Description
Bumps the version of `@newrelic/gatsby-theme-newrelic` to [1.2.1](https://github.com/newrelic/gatsby-theme-newrelic/releases/tag/v1.2.1) and uses the newly available `CodeBlock` and `MDXCodeBlock` components. Also includes a small fix to the background color for `<code>` tags used inline.

This PR resolves an issue where CodeBlock components were not rendering a `::selected` background color on the light theme. Since it wasn't an issue with the theme component, it was simply a matter of using that one.

* Closes #494

## Screenshot(s)
Screen Shot 2020-07-28 at 11 13 27 AM


src/components/GuideTile/GuideTile.js
```diff
@@ -16,33 +15,133 @@ const GuideTile = ({
...
+    base={Surface.BASE.PRIMARY}
+    css={css`
+      display: grid;
+      grid-template-rows: auto auto 1fr auto;
+      border-radius: 0.25rem;
+      position: relative;
+      padding: 1rem;
+      transition: all 0.15s ease-out;
+
+      &:hover {
```
jerelmiller, line 34: Rather than setting hover styles yourself here, you can use the `interactive` prop on the `Surface`. Using that prop will ensure consistency with other surfaces on all sites.

src/components/GuideTile/GuideTile.js
```diff
@@ -16,33 +15,133 @@ const GuideTile = ({
...
+      ${featured &&
+      `
+          border-color: var(--color-white);
+
+          .dark-mode & {
+            border-color: var(--color-dark-100);
+            background-color: var(--color-dark-100);
+          }
+
+          &:hover {
```
jerelmiller, line 55: Rather than unsetting the hover style for featured tiles, you can use the `interactive` prop to conditionally apply them. `interactive={!featured}`

src/components/GuideTile/GuideTile.js
```diff
@@ -16,33 +15,133 @@ const GuideTile = ({
...
+        h2 {
+          margin-top: 1rem;
+        }
+      `}
+
+      ${featured &&
+      `
+          border-color: var(--color-white);
+
+          .dark-mode & {
```
jerelmiller, line 50: We should try and leave the border/background styles to the `Surface`, since its goal is to handle the box shadow/border/background on surfaces rendered "above" other content. 

In this case, it looks like you're setting these background/border colors based on whether these are rendered on the featured section. Rather than conditionally applying styles, apply a different `base` to the surface. In this case, we want to render the surface on a "secondary" base when it's a featured tile. Remove these styles in favor of: `base={featured ? Surface.BASE.SECONDARY : Surface.BASE.PRIMARY}`

package.json
```diff
@@ -7,7 +7,7 @@
     "@emotion/styled": "^10.0.27",
     "@mdx-js/mdx": "^1.6.6",
     "@mdx-js/react": "^1.6.6",
-    "@newrelic/gatsby-theme-newrelic": "^1.1.10",
+    "@newrelic/gatsby-theme-newrelic": "^1.2.1",
```
jerelmiller, line 10: FYI: `1.2.2`  was published and includes the dark mode box shadow.

jerelmiller: ~~1.2.2~~ 1.2.3

PR #498, opened 2020-07-29
Use multi-episode player link.

thejonanshow:
## Description
This swaps out the player URL so multiple episodes will appear; the original embed was using the single episode player.

## Reviewer Notes
The player should automatically show new episodes after this gets deployed.

## Screenshot
Screen Shot 2020-07-29 at 9 51 54 AM


thejonanshow:
I apologize for not getting this into the original PR, I didn't realize there were different embed links.

PR #501, opened 2020-07-30
Index Page Context Updates

zstix:
## Description
Updates to the index page copy / button.

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -109,7 +109,7 @@ const IndexPage = ({ data, pageContext }) => {
               <Button
                 as={ExternalLink}
                 variant={Button.VARIANT.PRIMARY}
-                href="https://newrelic.com/signup/?utm_source=developer-site"
+                href="https://newrelic.com/signup?utm_source=developer-site"
               >
                 Create a free account
               </Button>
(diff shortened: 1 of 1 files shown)


src/pages/index.js
```diff
@@ -108,9 +109,9 @@ const IndexPage = ({ data, pageContext }) => {
               <Button
                 as={ExternalLink}
                 variant={Button.VARIANT.PRIMARY}
-                href="https://newrelic.com/signup?partner=Developer+Edition"
+                href="https://newrelic.com/signup/?utm_source=developer-site"
```
jerelmiller, line 112: It looks like the `/signup/` page (with the trailing `/`) doesn't work. Can you remove the trailing `/`?

zstix: Ah, copied and pasted from the doc. Kinda surprised that doesn't work. Either way, updating it now (via suggestion)
```suggestion
                href="https://newrelic.com/signup?utm_source=developer-site"
```

jerelmiller: Ya not sure why it doesn't work. I happened to see something mentioned in one of the slack channels.

jerelmiller:
The PR preview built. Not sure why its stuck in pending state here. Going to go ahead and merge :)

PR #502, opened 2020-07-30
Right rail content

### LINKED ISSUE #476: [Related Content] Create "right rail" component (UI only)
## Description
We need to create a component that will eventually house the related content. This component should be displayed to the right of the page content. At the time of writing, the design treatment for this is still in flux, but the current thinking is that this will sit to the right of the content but,a s the page scrolls up, the right rail will get locked to the top of the page (sticky).

_NOTE: this ticket is only for the markup / styles._

## Helpful Links
N/A

## Screenshots
**Work-in-Progress Wireframes**
unnamed

jerelmiller:
Closes #476 

## Description
Adds a "right rail" component that houses the area to contribute to the project. The right rail only shows up on guide pages.

## Screenshot(s)
localhost_8000_collect-data_collect-data-from-any-source (1)


src/pages/index.js
```diff
@@ -74,170 +53,183 @@ const IndexPage = ({ data, pageContext }) => {
 
   return (
     <PageContext.Provider value={pageContext}>
-      <>
-        <SEO />
-        <PageTitle>Observability for every developer</PageTitle>
+      <SEO />
+      <PageLayout type={PageLayout.TYPE.SINGLE_COLUMN}>
+        <PageLayout.Header title="Observability for every developer" />
```
zstix, line 58: This adds an underline to the title on the homepage, which doesn't currently exist. We might want to add a lil `css` override for this.

zstix: I'm actually not seeing the underline for most of our pages. We may just want to remove this an add it where applicable.

jerelmiller: I added the underline intentionally on every page, including the homepage, so that every page had a consistent page title treatment. I might need to chat with @djsauble one more time and get his thoughts 😄

jerelmiller:
> Also, would it make sense to move `src/layout/MainLayout.js` to the `src/components/PageLayout` directory?

Not quite. I can understand the confusion, which is why I had such a hard time trying to name the `PageLayout` component. The `MainLayout` is meant to represent the layout of the entire website (global header, side nav, main content, footer). The `PageLayout` is meant to represent the layout only of the content area (i.e. the "page" content). Right now this consists of 2 layouts: single column (overview pages, home page, etc) and related content (2 column with content area and the related content area).

We are going to want to add that related content area to more pages over time, so the thinking was to try and make this easy when we come back to this for other pages. Ideally its a switch of the `type` prop to the `PageLayout`, and the addition of the `PageLayout.RelatedContent` to the page.

From a developer point of view, I'd love to get your thoughts on this API and by all means, if you have a better name that `PageLayout`, I'm all for it. I'm not totally in love with it, but its the best I had 😄 .

jerelmiller:
@djsauble I think we need to figure out how we can make the content area bigger. That or maybe we need a better breakpoint for when the 2-column layout switches to a single column in the content area.

jerelmiller:
@zstix After talking with @djsauble we decided to keep the underline on every page to maintain consistency with the separation of the page title and the content. I've also pushed up some changes to 1) widen the layout by 100px to help those with super-wide monitors and 2) update the breakpoint that switches the side by side content to single column. That should help people with mid-size screens see both the related content and main content area without making it feel squished. Take a look and see if this works a bit better for you!

PR #550, opened 2020-08-06
ran npm audit fix on minor security warnings

jpvajda:
## Description
When running npm install on a recent git pull, I saw over 5000 minor security warnings, so i ran npm audit fix to correct it.

## Reviewer Notes

I wanted to submit this as a separate PR in case we don't want to update the lock file.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1615,9 +1615,9 @@
       }
     },
     "@babel/standalone": {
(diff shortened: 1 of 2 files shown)


jpvajda:
After speaking with @jerelmiller  I'm going to close this out as it got complex and into some weird dependencies issues with dot-prop

PR #585, opened 2020-08-12
added a br tag under the intro

jpvajda:
## Description
no space looked bad so i fixed it.


jerelmiller:
Should we consider adding a margin on the [`Intro` component](https://github.com/newrelic/developer-website/blob/main/src/components/Intro.module.scss) instead? I suspect other pages might have this problem as well.

jpvajda:
@jerelmiller we definitely can, but I guess we'll have to consider if we add it the component these pages will need the <br> tags removed.

PR #594, opened 2020-08-13
Source related resource content from Swiftype

### LINKED ISSUE #584: Filter out pages defined via frontmatter from Swiftype related content results
## Description

Because we are mixing related resources from frontmatter and Swiftype, there is a potential that Swiftype would return results for a page defined in frontmatter. To avoid displaying a duplicate result, we need to tell Swiftype to filter out those URLs when searching. We can leverage the [Swiftype Search API filtering](https://swiftype.com/documentation/site-search/searching/filtering) mechanism to do this for us.

You can see an example of this in the [search-research branch](https://github.com/newrelic/developer-website/blob/search-research/src/components/RelatedContentModules/Swiftype.js#L51-L57) that filters out the current page from results.

### LINKED ISSUE #589: Update the GraphQL schema to include related pages
## Description

We want to update the GraphQL schema generated by Gatsby to include related pages. At the very least, this new type in the GraphQL system will need to include both a `title` and `url` field. We will want to leverage the [Gatsby Node APIs](https://www.gatsbyjs.com/docs/node-apis) to do so.

The data should resolve from a JSON file written to by the cron task from #582 . 

## Useful links

* [Gatsby Node APIs](https://www.gatsbyjs.com/docs/node-apis)

jerelmiller:
Closes #584 
Closes #589 

## Description

Adds the ability to source data from Swiftype to populate the related resource content. This adds the GraphQL nodes needed to fetch from the site as well as the flags to tell it whether to regenerate results from Swiftype or read from the local JSON file. This also ensures results from the frontmatter defined resources are filtered out from the search call.


gatsby-config.js
```diff
@@ -42,6 +44,51 @@ module.exports = {
         },
       },
     },
+    {
+      resolve: 'gatsby-source-swiftype',
+      options: {
+        file: `${__dirname}/src/data/related-pages.json`,
+        refetch: Boolean(process.env.BUILD_RELATED_CONTENT),
```
caylahamann, line 51: Is this for caching reasons?

jerelmiller: Yes. By default, the data is resolved from the `src/data/related-pages.json` file. When we need to refresh that data and make new requests to Swiftype, setting the `BUILD_RELATED_CONTENT` environment variable will instruct the plugin to go make new requests to Swiftype and save them in the JSON file.

PR #596, opened 2020-08-13
Instruqt k8s operator

polfliet:
## Description
Pilot for embedding Instruqt tutorials. This is a quick tutorial following the steps to install and use the k8s operator.
The user will see a terminal screen with instructions to follow. The content of this tutorial can be updated in the Instruqt platform and is hosted here: https://github.com/polfliet/instruqt/tree/master/nr-k8s-operator

## Reviewer Notes
It would be nice if we could show license key, account id, etc. for logged in users to offer them a better experience. Or we could see if it's possible to directly feed that into the Instruqt environment. 

## Screenshot(s)
Screenshot 2020-08-13 at 19 24 31


zstix:
We're working on adding better support for `iframe` tags in our MDX. Here's the PR for that work: #609 

Once that is merged, I'll give this a review 👀 an approval ✅ and a merge 🤿

PR #597, opened 2020-08-13
Related content in the right rail

### LINKED ISSUE #580: Show up to 5 links in the right rail
## Description

We want to show up to 5 links in the right rail component. These can be a mix of frontmatter-defined links and dynamic content links. Let's prioritize frontmatter-defined links first, then back fill results up to 5 links. For example, if there are 3 frontmatter-defined links, we want 2 search results from Swiftype for a total of 5. If there are 5 links defined in frontmatter, we should skip the request to Swiftype entirely for search results.

NOTE: This should consume the data from the GraphQL schema from #589 .

jerelmiller:
Closes #580 

## Description

Adds related resources from Swiftype to the right rail. This PR will back fill up to 5 links if there are frontmatter defined resources.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -1,6 +1,8 @@
 const path = require(`path`);
 const { execSync } = require('child_process');
 
+const MAX_RESULTS = 5;
+
 const getFileRelativePath = (absolutePath) =>
   absolutePath.replace(`${process.cwd()}/`, '');
 
@@ -17,6 +19,9 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
               path
               template
               redirects
+              resources {
+                url
+              }
             }
           }
         }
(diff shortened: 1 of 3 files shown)


jerelmiller:
DO NOT MERGE UNTIL THE BASE BRANCH HAS BEEN UPDATED TO `develop`

gatsby-node.js
```diff
@@ -53,6 +58,10 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
           frontmatter.template === 'OverviewTemplate'
             ? `${frontmatter.path}/*`
             : undefined,
+        relatedResourceLimit: Math.max(
```
caylahamann, line 61: Will this sent a request regardless of whether there are related resources to be requested? Shouldn't we not send out the request if the limit is 0?

jerelmiller: If by "request", you mean GraphQL request, then yes it will always execute. Since the GraphQL queries are static in Gatsby, unfortunately conditionally requesting fields from the schema are not really possible. This will still request data from the schema, but it means that if the limit is `0`, an empty array will be returned rather than the field being empty.

jerelmiller: One other thing to mention is that requests to Swiftype are only done at build time if the `BUILD_RELATED_CONTENT` environment variable is set, otherwise it just reads data from the `src/data/related-pages.json` file.

caylahamann: ahhhh okay! thank you for clarifying that.

PR #612, opened 2020-08-15
Attend event icon logic 

jpvajda:
## Description
Work to get the feather icon to appear from the attend event side navigation using an if statement. This isn't working yet, but it's super close so I'm handing it off to @jerelmiller  the "closer"

### DIFF
--- a/src/components/NavigationItems.js
+++ b/src/components/NavigationItems.js
@@ -80,7 +80,7 @@ const NavigationItems = ({
 };
 
 const NavIcon = ({ page }) => {
-  if (iconLibrary[page.displayName == 'Attend events']) {
+  if (iconLibrary[page.displayName === 'Attend events']) {
     return (
       <FeatherIcon
         className={styles.headerIcon}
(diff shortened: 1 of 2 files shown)


src/components/NavigationItems.js
```diff
@@ -16,8 +16,8 @@ const iconLibrary = {
   'Build apps': 'buildApps',
   'Automate workflows': 'automation',
   'Explore docs': 'developerDocs',
+  'Attend events': 'event',
```
jerelmiller, line 19: I somehow missed that this event icon for the new relic icon library got added. I like keeping that one, so lets do it!

src/components/NavigationItems.js
```diff
@@ -80,7 +80,14 @@ const NavigationItems = ({
...
 const NavIcon = ({ page }) => {
-  if (iconLibrary[page.displayName]) {
+  if (iconLibrary[page.displayName === 'Attend events']) {
+    return (
+      <FeatherIcon
+        className={styles.headerIcon}
+        name={iconLibrary[page.displayName]}
+      />
+    );
+  } else if (iconLibrary[page.displayName]) {
```
jerelmiller, line 90:
```suggestion
  if (iconLibrary[page.displayName]) {
```

Since we are using the arrow icon, we can revert this back to the way it was.

Issue #538, opened 2020-08-05
Provide documentation on this page for how to write to the NerdletStateContext


cirne:
There is no documentation on this page for how to _write_ to the `NerdletStateContext`. Presumably that ends up modifying the url params. Please presume I don't have a thorough understanding of React Context stuff, which feels more like React 202, not the core stuff I would expect most React developers to fully understand.

Missing documentation would be added here: https://developer.newrelic.com/components/nerdlet-state-context

jerelmiller:
Related #450

jpvajda:
Hello @cirne We'll close this one as duplicate and work from #450  Thank you for reporting this!

PR #621, opened 2020-08-17
Add the ability to provide supplemental API documentation

### LINKED ISSUE #450: Properly document the nerdlet.setUrlState method
### Description

Not all API methods are documented in the developer documentation. For instance, `nerdlet.setUrlState` method is not mentioned in the [developer documentation under platform API / nerdlet](https://developer.newrelic.com/apis/nerdlet ) but it’s part of the how to understand url state guide defined on [Discuss](https://discuss.newrelic.com/t/how-to-understand-nerdlet-url-state/97596)

### Steps to reproduce

1. Go to https://developer.newrelic.com/apis/nerdlet
2. Try to find any information on nerdlet.setUrlState
3. Go to https://discuss.newrelic.com/t/how-to-understand-nerdlet-url-state/97596
4. Try to see all the information on nerdlet.setUrlState

### Expected behavior

We properly document this method in the Platform API / nerdlet section of the developer site.

 It would be really nice to cross-link between [NerdletStateContext](https://developer.newrelic.com/components/nerdlet-state-context) and the [Nerdlet API](https://developer.newrelic.com/apis/nerdlet) since they are both intertwined, perhaps use the new RIGHT RAIL to highlight the page association.

### Additional context

We looked into updating the NR1 Core documentation for Platform API / nerdlet  as it's also not explained there to ensure the SDK Bundler can auto generate this method's documentation in the future. However that cannot be done at this moment, so this will be a manual change on the developer site to ensure this functionality is clearly document. 

Given we will do this manually we should ensure an auto update from the SDK Bundler won't overwrite this manual entry.

zstix:
## Description
The bundle that we use to generate documentation for components and APIs is missing information. One example is the `nerdlet.setUrlState` method. In order to help provide more complete documentation, I created a system for us to add additional documentation that we define. This (temporary) solution should allow us to add in any missing methods, typeDefs, or constants.

## Reviewer Notes
At the moment, I have just added the `nerdlet.setUrlState` method documentation (as the linked ticket suggests). Please let me know what you think of this setup / naming convention.

* Closes #450 

## Screenshot(s)
The first method in this screenshot is defined by us, the rest come from the bundle:
Screen Shot 2020-08-17 at 3 05 37 PM

### DIFF
--- a/src/data/additionalDocs.js
+++ b/src/data/additionalDocs.js
@@ -5,7 +5,7 @@ const additionalDocs = {
         name: 'nerdlet.setUrlState',
         description: `
 Updates the current nerdlet's URL state that can be read from
-\`NerdletStateContext.Consumer\`.
+[\`NerdletStateContext.Consumer\`](/components/nerdlet-state-context).
 
 If you wish to update nerdlet's state without persisting its value in the
 url, use React's built-in \`setState()\`.
--- a/src/data/additionalDocs.js
+++ b/src/data/additionalDocs.js
@@ -1,0 +1,51 @@
+const additionalDocs = {
+  nerdlet: {
+    methods: [
+      {
+        name: 'nerdlet.setUrlState',
+        description: `
+Updates the current nerdlet's URL state that can be read from
+\`NerdletStateContext.Consumer\`.
+
+If you wish to update nerdlet's state without persisting its value in the
+url, use React's built-in \`setState()\`.
+
+This method behaves like React's \`setState()\`, meaning that it performs a
+shallow merge between the current URL state and the provided state in the
+\`urlState\` parameter.
+
+If you wish to navigate without adding an entry to the browser history, set
+\`urlStateOptions.replaceHistory\` to \`true\`.
+        `.trim(),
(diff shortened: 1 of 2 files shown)


src/data/additionalDocs.js
```diff
@@ -0,0 +1,51 @@
+const additionalDocs = {
+  nerdlet: {
+    methods: [
+      {
+        name: 'nerdlet.setUrlState',
+        description: `
+Updates the current nerdlet's URL state that can be read from
+\`NerdletStateContext.Consumer\`.
```
jerelmiller, line 8: Could we link this to the documentation for `NerdletStateContext` here? Its one step toward solving an issue @cirne raised in #538 about how to use `NerdletStateContext`, granted I think we likely need to do a bit more to fully solve that documentation gap.

jerelmiller:
```suggestion
[\`NerdletStateContext.Consumer\`](/components/nerdlet-state-context).
```

I probably should have provided a suggestion to make it easier on you 😄

PR #623, opened 2020-08-17
added vscode ext and updated component guide

jpvajda:
## Description
- Added vscode ext to the right rail on several pages using url resources 
- added vscode ext tip section to several pages
- added a visual studio resource site tag
- updated component guide explaining how the right rail resources component works.

### DIFF
--- a/COMPONENT_GUIDE.md
+++ b/COMPONENT_GUIDE.md
@@ -252,3 +252,39 @@ Text
 
 </Caution>
 ```
+
+## Related Resources
+
+The related resources component is controlled by specific Frontmatter slugs that
+are defined on a page.
+
+By setting the Frontmatter for `resources` and `tags` you can control what is populated
+in this component. Please review the [Style Guide](STYLE_GUIDE) for further details
+on these slugs.
+
+### Maximum references
+
+This component allows a maximum of 5 resources.
+
+### Resource population logic
+
+If no resources or tags are available in the page Frontmatter the component will fallback
+to using the page title as the search query term but you will
+always have 5 results, assuming Swiftype can return 5 results for that query.
+
+### Order of priority
+
+The order of priority for populating content in this component is driven by:
+
+1. Any resources defined in the page Frontmatter.
+2. Any tags defined in the the page Frontmatter will return send a search query term and return results from Swifttype.
+3. Any Swifttype results that match the page title.
+
+### Resource site tags
+
+Resource site tags control the labels that appears below each resource URL. These
+can be found in the [Resources.js](./src/components/RelatedContentModules/Resources.js).
+If you are adding a resource that doesn't currently have a defined site tag, you will need to
+modify this file an add a new site tag accordingly.
+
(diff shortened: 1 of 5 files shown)


COMPONENT_GUIDE.md
```diff
@@ -252,3 +252,39 @@ Text
...
 </Caution>
 ```
+
+## Related Resources
+
+The related resources component is controlled by specific Frontmatter slugs that
+are defined on a page.
+
+By setting the Frontmatter for `resources` and `tags` you can control what is populated
+in this component. Please review the [Style Guide](STYLE_GUIDE) for further details
```
jerelmiller, line 262: It looks like this `Style Guide` link 404s. Perhaps you need a `.md` extension on the end of this?

COMPONENT_GUIDE.md
```diff
@@ -252,3 +252,39 @@ Text
 
 </Caution>
 ```
+
+## Related Resources
```
jerelmiller, line 256: I'm a bit hesitant to include this section in the `COMPONENT_GUIDE.md` document, only because this document documents components available _in the markdown_ itself. This is directly referencing _frontmatter_ specific information, so its not something you'd use or reference in the markdown directly.

That being said, maybe its ok that we do this, but maybe we should consider whether we should include other useful frontmatter specific details. It kinda feels caught in between the style guide doc and this one otherwise. This information is super useful, so I'm not advocating against not including any of it, just figuring out where best to put it 😄 

Thoughts?

COMPONENT_GUIDE.md
```diff
@@ -252,3 +252,39 @@ Text
...
+If no resources or tags are available in the page Frontmatter the component will fallback
+to using the page title as the search query term but you will
+always have 5 results, assuming Swiftype can return 5 results for that query.
+
+### Order of priority
+
+The order of priority for populating content in this component is driven by:
+
+1. Any resources defined in the page Frontmatter.
+2. Any tags defined in the the page Frontmatter will return send a search query term and return results from Swifttype.
```
jerelmiller, line 280:
```suggestion
2. Any tags defined in the the page Frontmatter will send a search query term and return results from Swiftype.
```

COMPONENT_GUIDE.md
```diff
@@ -252,3 +252,39 @@ Text
...
+3. Any Swifttype results that match the page title.
```
jerelmiller, line 281: This is sorta a 2b if you want to be totally accurate. We don't send the page title as a query unless there are no tags for the current page and we really only send a single query to Swiftype per page. Its pretty much just an if/else condition.

jerelmiller:
oops, didn't realize this already got merged. Anyways, maybe some feedback to take in the future.

jpvajda:
@jerelmiller i'll fix it now.. thanks!

COMPONENT_GUIDE.md
jpvajda: Well it's how the component works, and the style guide references how the frontmatter works. I think it's OK.

PR #641, opened 2020-08-19
fix: update CFP date and reg text

jaesius:
## Description
update CFP date and reg text


jerelmiller:
Looks like this may have already been fixed in a different branch. I'll go ahead and close this PR

Issue #618, opened 2020-08-15
Submit Dev Site to Gatsby Showcase


jpvajda:
we want to add our cool gatsby sites to the [Gastby Showcase](https://www.gatsbyjs.com/showcase/)

**AC**

Add [sites.yml](https://www.gatsbyjs.com/contributing/site-showcase-submissions/) file to both OSS and Dev sites 
[submit](https://www.gatsbyjs.org/contributing/site-showcase-submissions/) our sites.
Ping the Head of Product at Gatsby - Dustin Schau. Maybe they'll want to talk to us.
Maybe Gatsby would be interesting in blogging about our use case at New Relic about how we are using Gatsby.

jpvajda:
all of sudden their documentation on how to do this is gone, there was a little confusion on how to generate the  portfolio screen shot I need to look into..

 # this URL is used to generate a screenshot
  url: https://titleofthesite.com/portfolio

jpvajda:
I'm going to host a screen shot in the repo of the site and get this finished up today.

jpvajda:
After getting the PR submitted and passing the tests I was informed Gatsby has paused their showcase submissions and is reworking the process.

https://github.com/gatsbyjs/gatsby/pull/26442 

So we can wait to do this once they define the new process.

jpvajda:
The showcase seems to be on hold for a while so closing as won't do for now.

PR #647, opened 2020-08-20
Sites yaml for gastby showcase

jpvajda:
## Description
- includes the file for sites.yaml
- includes a new screenshot directory
- includes a screenshot 

If there are any related [GitHub Issues](https://github.com/newrelic/developer-website/issues) or JIRA tickets, add links to them here.

#618


sites.yml
```diff
@@ -0,0 +1,31 @@
+- title: New Relic Developer Site
+
+  # this is the URL that is linked from the showcase
+  main_url: https://developer.newrelic.com/
+
+  # this URL is used to generate a screenshot
+  url: https://github.com/newrelic/developer-website/screenshots
```
jpvajda, line 7: I was curious if this needs to reference the actual file in the path, the example they gave didn't require the file name.

jerelmiller:
Are we sure that this `sites.yml` file should live here and not the one in the `gatsby` repo itself? 

From what I understood, I think its this file that needs to be edited: https://github.com/gatsbyjs/gatsby/blob/master/docs/sites.yml

Then again, I could be wrong since I haven't actually done it yet 😄

jpvajda:
@jerelmiller Oh wait, I think you are right! It was weird, their showcase documentation just disappeared one day, as I think they did some domain combining of .org and .com. So yeah, I'll close this. I'm actually glad it works this way, as I thought if we hosted the file, was gastby pulling that in from some automated process or a plugin.

jpvajda:
@jerelmiller  Can we merge in this file and directory + screenshot? I'll need that for the image in the showcase. As for the sites.yml file, I was thinking we can keep it as a reference file with site. so there is a copy of it.

jerelmiller:
Ya I can do that

PR #685, opened 2020-08-25
Nerd days event page

jaesius:
## Description
Refactor code to create a MarketoForm component


src/components/MarketoForm/Form.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css } from '@emotion/core';
+import './marketo.scss';
+
+const Form = ({ title, id, children }) => (
```
jerelmiller, line 6: [nit] Since you've named the folder `MarketoForm` and have used the name `MarketoForm` when importing it, could you also name this component `MarketoForm` so that it shows up in React devtools as such?

src/components/MarketoForm/Form.js
```diff
@@ -0,0 +1,42 @@
...
+        css={css`
+          font-size: 20px;
+          font-weight: bold;
+          text-align: center;
+        `}
+      >
+        {title}
+      </p>
+      <form id={`mktoForm_${id}`} />
+      {children}
```
jerelmiller, line 31: It looks like you might not need the `children` prop. Is this something that could be removed from the component?

src/components/MarketoForm/Form.js
```diff
@@ -0,0 +1,42 @@
...
+        position: relative;
+        background-color: var(--primary-background-color);
+        border-radius: 5px;
+        border: 1px solid var(--border-color);
+        padding: 20px;
+      `}
+    >
+      <p
+        css={css`
+          font-size: 20px;
```
jerelmiller, line 23: [nit] We've been trying to stick to `rem`s for font size and padding. Could you update this to `1.25rem`?

src/hooks/useMarketForm.js
```diff
@@ -0,0 +1,36 @@
...
+        }
+      }, 250);
+    };
+    const script = document.createElement('script');
+    script.src = 'https://marketo.clearbit.com/assets/v1/marketo/forms.js';
+    script.async = true;
+    script.setAttribute('data-clearbit-publishable-key', pubishableKey);
+    script.onerror = function () {
+      // eslint-disable-next-line no-console
+      console.log('Clearbit Form JS unable to load');
```
jerelmiller, line 24: We do have New Relic browser installed on the dev site. Would it be better to maybe send an event through that? That would make it easier to detect when the form is unable to load so we can look into this if we see it happening a lot.

src/hooks/useMarketForm.js
```diff
@@ -0,0 +1,36 @@
+import { useEffect } from 'react';
+
+const useMarketoForm = (munchkinId, id, pubishableKey) =>
```
jerelmiller, line 3:
```suggestion
const useMarketoForm = (munchkinId, id, publishableKey) =>
```

Looks like maybe this was a typo

src/hooks/useMarketForm.js
```diff
@@ -0,0 +1,36 @@
+import { useEffect } from 'react';
```
jerelmiller, line 1: It looks like the name of the file is `useMarketForm`, but the hook is named `useMarketoForm`. I assume thats a typo. Could you update the name of the file to match?

src/hooks/useMarketForm.js
```diff
@@ -0,0 +1,36 @@
...
+
+const useMarketoForm = (munchkinId, id, pubishableKey) =>
+  useEffect(() => {
+    window.MktoForms2.loadForm('//app-abj.marketo.com', munchkinId, id);
+
+    const pollForDefinition = function (scope, varname, callback) {
```
jerelmiller, line 7:
```suggestion
    const pollForDefinition = (scope, varname, callback) => {
```

[nit] we typically use the anonymous arrow function syntax and tend to avoid the `function` keyword when possible.

src/hooks/useMarketForm.js
```diff
@@ -0,0 +1,36 @@
...
+      if (typeof scope[varname] !== 'undefined') {
+        return callback();
+      }
+      const interval = setInterval(function () {
```
jerelmiller, line 11:
```suggestion
      const interval = setInterval(() => {
```

src/hooks/useMarketForm.js
```diff
@@ -0,0 +1,36 @@
...
+          clearInterval(interval);
+          callback();
+        }
+      }, 250);
+    };
+    const script = document.createElement('script');
+    script.src = 'https://marketo.clearbit.com/assets/v1/marketo/forms.js';
+    script.async = true;
+    script.setAttribute('data-clearbit-publishable-key', pubishableKey);
+    script.onerror = function () {
```
jerelmiller, line 22:
```suggestion
    script.onerror = () => {
```

src/hooks/useMarketForm.js
```diff
@@ -0,0 +1,36 @@
...
+      // eslint-disable-next-line no-console
+      console.log('Clearbit Form JS unable to load');
+      pollForDefinition(window, 'MktoForms2', function () {
```
jerelmiller, line 25:
```suggestion
      pollForDefinition(window, 'MktoForms2', () => {
```

src/hooks/useMarketForm.js
```diff
@@ -0,0 +1,36 @@
...
+        window.MktoForms2.whenReady(function (form) {
```
jerelmiller, line 26:
```suggestion
        window.MktoForms2.whenReady((form) => {
```

src/pages/nerd-days.js
```diff
@@ -1,47 +1,15 @@
...
-        window.MktoForms2.whenReady(function (form) {
-          form.setValues({
-            clearbitFormStatus: 'Clearbit Form JS unable to load',
-          });
-        });
-      });
-    };
-    document.body.append(script);
-  }, []);
+  useMarketoForm('412-MZS-894', 3525, 'pk_4d10daa544de6f993a9a9ce002ccd1c6');
```
jerelmiller, line 12: Looking at this, I think it might make sense to move this hook call inside the `MarketoForm` component and pull these values from its props. This would allow us to use a single component for future events without needing to remember to use both the hook and the component.

```js
const MarketoForm = ({ munchkinId, id, publishableKey }) => {
  useMarketoForm(muchkinId, id, publishableKey)

  return // ...
}

// usage
<MarketoForm 
  id={3525} 
  munchkinId="412-MZS-894" 
  publishableKey="pk_4d10daa544de6f993a9a9ce002ccd1c6" 
/>
```

One nice side effect of that is that you don't have to repeat the `id` between the hook and the component, which might be error prone.

Thoughts?

jaesius: Great suggestion! Being able to use a single component would be best for lowering the number of errors in the future.

src/hooks/useMarketForm.js
jaesius: Good idea, I'll create an issue for this, being that I have a couple of questions for you about adding it to the NR browser.

PR #688, opened 2020-08-25
Add files via upload 

veextee:
Adding all things open image, so we can use it for event landing page.


jerelmiller:
Looks like we got the image as part of #698 so I'm going to close this PR

PR #695, opened 2020-08-28
Nerd days event page

jaesius:
## Description
Add EMEA and APJ event page


src/pages/apj-nerd-days.js
```diff
@@ -0,0 +1,74 @@
...
+          <section className={cx(styles.section, styles.twoColumnAlt)}>
+            <div>
+              <img
+                className={styles.img}
+                src={nerdDays}
+                alt="nerd days header"
+              />
+              <p>
+                Nerd Days is a <b>FREE</b> engineering conference that kicks off
+                October 22 <em>(Dates vary by region)</em>. Focused on building
```
caylahamann, line 26: Since the nerd day pages seem to be identical except for varying the date and region, it might be worth it to pull this out into a template or component and then just render the template for each page in here with the props passed in.

src/components/MarketoForm/MarketoForm.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
```
caylahamann, line 1: would you be able to check this out in a new branch and new PR? It seems that because we're still on the same branch, it's registering these changes as new ones to merged into master even though those have already been merged in

src/pages/apj-nerd-days.js
jaesius: The content and structure will vary per page. They are just the same for now.

src/components/MarketoForm/MarketoForm.js
jaesius: Will do!

jerelmiller:
Closing since this is a duplicate of #696

PR #696, opened 2020-08-28
Nerd days global pages

jaesius:
## Description
Add global event pages


src/pages/emea-nerd-days.js
```diff
@@ -0,0 +1,74 @@
...
+                  {' '}
+                  check out our guide.
+                </a>
+              </p>
+            </div>
+            <MarketoForm
+              id={3525}
+              title="REGISTER FOR NERD DAYS | EMEA"
+              munchkinId="412-MZS-894"
+              publishableKey="pk_4d10daa544de6f993a9a9ce002ccd1c6"
```
jerelmiller, line 64: Just to double check before I approve/merge, are these keys supposed to be unique or the same as the apj nerd days event? If they are all the same form, great! Just wanted to verify that was the case 🙂

jaesius: Yes, they are the same. It seems the URL is what matters not the keys

jerelmiller: Perfect!

PR #714, opened 2020-09-03
Split.io integration

### LINKED ISSUE #645: Add the edit and create an issue icons back to global header but put search as the first icon:  9/2
The theory is, that we want to call out these contributor actions throughout the site experience and the best place to do that is in the header.  So we should add those icons back to the global header in this order:

Search | Dark mode toggle | edit page | create issue 

When making this change the OSS site should also be updated accordingly as it uses the global header.

### LINKED ISSUE #655: Set up Split.IO
Determine the following:

In order to support our experiment with using split.io for A/B Testing I _think_ we may need to install that Gatsby plugin for this functionality. 

https://www.gatsbyjs.com/plugins/gatsby-plugin-splitio/

We should looking this plugin and how it works with Split.io

jerelmiller:
Closes #645 
Closes #655 

## Description

Sets up split integration for the site. Will also conditionally add the github buttons to the header based on the treatments of the split.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -18,6 +18,21 @@ module.exports = {
           contentPadding: '2rem',
           maxWidth: '1700px',
         },
+        splitio: {
+          core: {
+            authorizationKey: process.env.SPLITIO_AUTH_KEY,
+          },
+          env: {
+            development: {
+              features: {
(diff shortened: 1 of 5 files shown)


caylahamann (approved):
looooks all good to me 👍

PR #725, opened 2020-09-08
Announcement Banner Fixes

zstix:
## Description
This addresses a few issues that are specific to the developer site when it comes to using the _Announcement Banner_:
* Fixes an issue where Gatsby would try to create a new page for an announcement MDX file
* Fixes an issue where the sidebar was appearing _under_ the header when a banner was present

## Reviewer Notes
I added a hook called `useResizeObserver` that uses [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) under the hood. This may be helpful in other locations. At the moment, it's just set up to work with height, but we can make it more generalized if we want.

N/A

## Screenshot(s)
**With a Banner:**
Screen Shot 2020-09-08 at 2 57 05 PM

**Without a Banner:**
Screen Shot 2020-09-08 at 2 57 15 PM

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -11,7 +11,10 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
 
   const result = await graphql(`
     {
-      allMdx(limit: 1000) {
+      allMdx(
+        limit: 1000
+        filter: { fileAbsolutePath: { regex: "/src/markdown-pages/" } }
+      ) {
         edges {
           node {
             fileAbsolutePath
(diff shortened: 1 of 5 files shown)


jerelmiller:
It looks like we may need to polyfill the `ResizeObserver` when building.

zstix:
I fixed the build issue with `ResizeObserver` but it looks like the build fails on some new functionality that comes from the `AnnouncementBanner` in the theme. Going to work on a fix there and then bump the version in this branch.

caylahamann (approved):

PR #741, opened 2020-09-11
Nerd days updates

jaesius:
## Description
Added registration section to the main Nerd Days landing page to bypass needing a second US page, updated APJ/EMEA date, adding date and times to us/apj/emea registration pages.


jpvajda (commented):
I scanned this and it LGTM... but maybe a dev should check it as well.

zstix (approved):

PR #768, opened 2020-09-17
Swap out URL

veextee:
I'd prefer for us to put the nominee page, so we can show all the categories we've been nominated for. https://devportalawards.org/nominees/2020/new-relic
image


zstix (approved):
Looks good to me!

PR #775, opened 2020-09-18
Fix marketo content blocking issue.

curious-attempt-bunny:
## Description
With Content Blocking enabled in Firefox the page vanishes in prod, and looks like this locally:
Pasted_Image_9_18_20__1_06_PM

## Reviewer Notes
Use Firefox, enabled content blocking. Very before and after.

This is a JIRA for this. Not sure what it is.

## Screenshot(s)
I can't see amplify's failure, this is what it looks like for me after the fix:
image

### DIFF
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -42,9 +42,11 @@ const onInitialClientRender = () => {
   }
 
   // eslint-disable-next-line no-undef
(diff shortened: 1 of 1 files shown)


curious-attempt-bunny:
I signed the CLA and clicked recheck. 🤷 Seems odd to need it as I believe I'm a newrelic org member.

PR #793, opened 2020-09-27
NR1 SDK data source in GraphQL

jerelmiller:
## Description

Moves the data for the NR1 SDK into GraphQL so that we can query for the docs at build time. This is much more SEO friendly as it allows us to generate static HTML pages during build with all of the SDK documentation in it.

Note this PR is only an attempt at creating the plugin necessary to query the data. This PR does not yet update the component and API documentation to use it. This PR is already pretty big, so an there will be a followup PR to pull data from GraphQL on the documentation pages.


jerelmiller (commented):
Some notes for others reviewing this PR

caylahamann (commented):
This all looks great to me (sorry it took me a while to get through, had to read up a bit on gatsby-plugins). The only question that I have is are we going to add in the additional doc stuff in this PR and are we going to make the loading into the head of the page async?

jerelmiller:
> are we going to add in the additional doc stuff in this PR and are we going to make the loading into the head of the page async?

I wasn't planning on it. I will followup with an additional PR for this stuff. This is just hooking up the data to GraphQL, but isn't consumed anywhere. Seeing as this PR is already big enough, I figured those things were easier to review in a later PR.

caylahamann (approved):

PR #807, opened 2020-10-01
Commit changes to related content on the main branch

jerelmiller:
## Description

Updates the GitHub action, that triggers the updates to related content, to 
commit to `main` instead of `develop`. This ensures the site is redeployed when 
these changes occur without needing a team member to open a PR from `develop` to
`main`

### DIFF
--- a/.github/workflows/fetch-related-content.yml
+++ b/.github/workflows/fetch-related-content.yml
@@ -15,6 +15,8 @@ jobs:
     steps:
       - name: Checkout repo
         uses: actions/checkout@v2
+        with:
+          ref: main
 
       - name: Setup node.js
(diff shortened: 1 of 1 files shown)


.github/workflows/fetch-related-content.yml
```diff
@@ -47,50 +49,12 @@ jobs:
           git commit -m 'chore(related-content): updated related content data'
           echo "::set-output name=commit::true"
 
-      - name: Temporarily disable "required_pull_request_reviews" branch protection
```
jerelmiller, line 50: Removed these because we do not currently have branch protection on the main branch.

roadlittledawn (approved):
Straightforward enough :shipit:

PR #817, opened 2020-10-05
Removed extra s from users'

jpvajda:
there was an extra s on users' so I removed it based on this:

possessive singular: user's
possessive plural: users'


jerelmiller (approved):
Good find!

PR #889, opened 2020-10-23
Update callouts to use the theme callouts

jerelmiller:
## Description

Update the all callouts to use the callouts from the theme. This removes the old `<Tip />`, `<Caution />`, and `<Important />` components.

## Screenshots

Screen Shot 2020-10-23 at 3 21 22 PM


roadlittledawn:
[nit] should we adjust the margin for the <h4> for the callout heading so the text isn't as offset from top as it is from bottom?

jerelmiller:
@roadlittledawn Yes! We need to fix that in the theme though, and unfortunately that component doesn't take a `className` prop. I am just about to go do some work over in the theme!

roadlittledawn:
ah gotcha! looks good to me then. will approve now

roadlittledawn (approved):
:shipit:

PR #894, opened 2020-10-26
Switch to yarn

jerelmiller:
## Description

Since we are already using `yarn` for the gatsby theme and the docs site, this PR switches the dev site to `yarn` to better align with the sites our team owns.

This PR also addresses a couple other issues:

* Temporarily disables branch protection for `main`. Previously we did not have any branch protection on `main`, but we would like to. This ensures we can do so safely.
* Ensures semantic release runs for the `develop` branch so that versions + changelog are generated again.


caylahamann (approved):
This all looks good to me, just need to make sure we make a big annoucement to the docs writers that we have made this change :)

jerelmiller:
Totally :)

Issue #888, opened 2020-10-23
Add oss and devsite domains to google search console


jpvajda:
Google Analytics is user-oriented, providing data related to those who visit and interact with your website. Google Search Console, on the other hand, is search-engine focused, providing tools and insights that can help site owners improve visibility and presence in the SERPs.

We currently don't have our domains verified in Google Search console which we should do to get search analytics. 

Google has 4 ways to validate your site We just need to decide which approach to use and implement that approach

<img width="610" alt="Screen Shot 2020-10-23 at 1 21 13 PM" src="https://user-images.githubusercontent.com/1395158/97052105-a461e700-153d-11eb-9c32-e988b24ce4f3.png">

jpvajda:
I tried all those verification methods, and they don't validate we are looking into it and think that the Google UA tag is not being fired due to our cookie acceptance component, so Google isn't able to validate it.  @jerelmiller thinks  its because we don’t install the tracking code until after the user agrees to accept the cookies

caylahamann:
For this we might have rewrite the way we do cookie consent in our site, so I'm going to read through the google analytic docs to see if we're able to correctly configure this.

jpvajda:
We've tried several approaches to this and none have worked unfornately so we are going to go with the simplest which is the HTML meta tag. 

@caylahamann here are the instructions, when I return I can validate this in Google Search Console. OR @zstix may have access to Search Console and he can validate it. :) 

To validate you just click on the menu in the right corner of the console:

![Screen Shot 2020-10-30 at 6 09 55 PM](https://user-images.githubusercontent.com/1395158/97766216-1eb0df00-1adb-11eb-91ff-62cf476e15d7.png)


1. Copy the meta tag below, and paste it into your site's home page. It should go in the <head> section, before the first <body> section.

**developer site**
<meta name="google-site-verification" content="He_vizRXYX_mUhwBe3BmyaMxNnVRAZbq_Jtm2A0e4WY" />

**oss site**
<meta name="google-site-verification" content="He_vizRXYX_mUhwBe3BmyaMxNnVRAZbq_Jtm2A0e4WY" />

roadlittledawn:
FWIW, we're verified with Google Search Console. not sure if there's anything to do except give additional peoples access?

jpvajda:
this is done. :)

PR #914, opened 2020-10-29
Google analytics refactor 

caylahamann:
## Description
Uses gatsby-plugin-gdpr-tracking and `gtag.js` to add Google Analytics scripts into the head of our site. It sets google analytics tracking to false on start and then by setting `newrelic-gdpr-consent` to true in cookies and invoking `window.trackGoogleAnalytics()`, it will start tracking without page reloads. 

## Reviewer Notes
To test out in development, you can set in options the following configuration with debug set to `true` and environments to include `'development'`: 
```js 
options: {
        debug: true,
        googleAnalytics: {
          .....
        },
        environments: ['production', 'development'],
      },
``` 
In the console, you'll be able to see logging that shows when the tracking starts. This adds the google tag to our page, but it doesn't fire page views or tracks until the user has consented. This should allow us to add the dev site to the google console. 

Partially addresses #888


package.json
```diff
@@ -70,10 +68,12 @@
     "eslint-plugin-jsx-a11y": "^6.3.1",
     "eslint-plugin-markdown": "^1.0.2",
     "eslint-plugin-react-hooks": "^4.2.0",
+    "gatsby-plugin-gdpr-tracking": "^1.2.2",
     "husky": "^4.2.5",
     "identity-obj-proxy": "^3.0.0",
     "jest": "^26.0.1",
     "jest-emotion": "^10.0.32",
+    "npm": "^6.14.8",
```
jerelmiller, line 76: Does this need to be here?

src/layouts/MainLayout.js
```diff
@@ -46,15 +46,18 @@ const MainLayout = ({ children }) => {
...
     setIsMobileNavOpen(false);
   }, [location.pathname]);
 
+  useEffect(() => {
+    if (cookieConsent && previousCookieConsent === false) {
+      window.trackGoogleAnalytics();
+    }
+  }, [cookieConsent, previousCookieConsent]);
+
+  const previousCookieConsent = usePrevious(cookieConsent);
```
jerelmiller, line 59: I _think_ this may have trouble building since you're declaring this variable after you're using it on line 57. Could you move this above the previous `useEffect`?

src/layouts/MainLayout.js
```diff
@@ -46,15 +46,18 @@ const MainLayout = ({ children }) => {
...
-    const consentValue = Cookies.get(gdprConsentCookieName) === 'true';
-    consentValue && setCookieConsent(true);
-  }, []);
-
   useEffect(() => {
     setIsMobileNavOpen(false);
   }, [location.pathname]);
 
+  useEffect(() => {
+    if (cookieConsent && previousCookieConsent === false) {
```
jerelmiller, line 54: If the user has accepted cookies, does it automatically start tracking without needing to call `window.trackGoogleAnalytics`?

jerelmiller (changes requested):
Spotted a couple minor things

src/layouts/MainLayout.js
caylahamann: yes! so it runs on route change, but if the user has already accepted the cookies on a previous visit then it doesn't need to call `window.trackGoogleAnalytics`

package.json
caylahamann: nope haha not sure how that got there

src/layouts/MainLayout.js
caylahamann: yes!

src/layouts/MainLayout.js
jerelmiller: Just to also clarify, if you reload a page after accepting cookies, does it record that as a hit without needing to change routes first?

caylahamann: yes! it does record as a hit without needing to change routes

jerelmiller (approved):

PR #960, opened 2020-11-17
image upload

veextee:
adding image needed to update homepage to include countdown to AWS re:Invent


jerelmiller (approved):
We'll need to make a small tweak to the homepage to render this image, but this works great!

PR #980, opened 2020-11-24
chore: api content updates

jpvajda:
## Description
- Adds navigation for additional APIs under the try our APIs section
- Adds an API index page to introduce our different APIs better
- Adds information to the build a hello world guide from the old developer site about proxy configuration and web pack extending per Miguel's request.
- Updates the Hello world app guide to explain the icon.png better. 

## Related Issue(s) / Ticket(s)

https://github.com/newrelic/developer-website/issues/959

## Screenshot(s)
Screen Shot 2020-11-24 at 2 55 06 PM

PR #982, opened 2020-11-28
fix: remove duplicate section at end of hello-world-app docs

tyreer:
## Description
Howdy! Just following this guide and noticed the duplicated section at the end. Looks like the Related Info === Summary and I'd imagine we want Summary as the header there.

- https://developer.newrelic.com/build-apps/build-hello-world-app

## Screenshot(s)
Screen Shot 2020-11-28 at 2 42 28 PM


jpvajda:
I fixed this in #980

PR #985, opened 2020-12-01
Fix failing builds

zstix:
## Descriptions
Our builds were failing once we bumped the theme to `v1.15.0`. This was because our builds expecting a `SPLITIO_AUTH_KEY` to be defined in our `gatsby-config.js`.


zstix:
We've found a way to resolve this by updating the theme itself, not each site. Going to close this PR and create another one that bumps the theme version.

PR #999, opened 2020-12-03
Bump version for New Relic theme

zstix:
## Description
Does what it says on the tin. This should resolve an issue we are having with the examples on the component docs pages not rendering.

## Reviewer Notes
There was a specific plugin commit targeted previously. I forget why this was the case. Let me know if we need to work around something.


package.json
```diff
@@ -20,7 +20,7 @@
     "gatsby-plugin-manifest": "^2.4.35",
     "gatsby-plugin-mdx": "^1.2.47",
     "gatsby-plugin-meta-redirect": "^1.1.1",
-    "gatsby-plugin-newrelic": "newrelic/gatsby-plugin-newrelic.git#5cd42e832f0588764dc332ce7ebd05794de5106a",
+    "gatsby-plugin-newrelic": "^1.0.5",
```
jerelmiller, line 23: Since this is installed from the theme, do we need this here?

zstix: I guess not? I'll remove it

jerelmiller (approved):

PR #1022, opened 2020-12-10
Fix names of Slack and PagerDuty

austin-schaefer:
## Description

The page isn't using the proper capitalization for these vendors


jerelmiller (approved):
💯

PR #1032, opened 2020-12-15
Change quote to callout

alexronquillo:


jerelmiller (approved):
Thanks for updating that!

PR #1045, opened 2020-12-18
update button for terraform guide

veextee:
current button directs to annemunition promotion from re:invent - need to update to the terraform guide.


jerelmiller (commented):
Great catch @veextee! I always love seeing PRs from you 😄 

I added 2 minor suggestions to this edit. It looks like using the `ExternalLink` will open in a new tab, which is a bit weird when navigating within the site. Other than that, this looks great!

jerelmiller (approved):

PR #1071, opened 2021-01-27
Relicans

aaronbassett:
## Description

Add New Relic Developer Relations team page

## Reviewer Notes

There is a new page listed in the side nav under "The Relicans"


src/data/sidenav.json
```diff
@@ -597,6 +597,10 @@
       }
     ]
   },
+  {
+    "displayName": "The Relicans",
```
jpvajda, line 601: Though this is proposed to be a primary nav item, eventually it may be better suited under a section like Developer Relations, where we could put this content and our developer events. I'm ok with it being in the primary nav for now, but just something to consider.

src/pages/relicans.js
```diff
@@ -0,0 +1,347 @@
...
+                <li>
+                  To invite you to hang out with us for gaming, writing code, playing music, just having fun.
+                </li>
+              </ul>
+              
+            </div>
+            <iframe width="560" height="315" src="https://www.youtube.com/embed/OlW3g631ero" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+          </section>
+
+          <h2 style={{ textAlign: 'center', marginBottom: 30, background: '#fff'}}>Meet the Relicans</h2>
```
jpvajda, line 335: The treatment on this H2 in dark mode seems off and doesn't follow our pattern for dark mode, I'd suggest just following how we toggle font colors from dark to light. In dark mode all headers font colors are changed to color: rgb(232, 234, 234)

Screen Shot 2021-01-27 at 9 04 25 AM

src/pages/relicans.js
```diff
@@ -0,0 +1,347 @@
...
+import dannyAvatar from '../images/relicans/danny-ramos.jpg';
+import jonanAvatar from '../images/relicans/jonan-scheffler.jpg';
+import kirkAvatar from '../images/relicans/kirk-haines.jpg';
+import miaAvatar from '../images/relicans/mia-moore.jpg';
+import pachiAvatar from '../images/relicans/pachi-carlson.jpg';
+import rachaelAvatar from '../images/relicans/rachael-wright-munn.jpg';
+
+const RelicansPage = () => {
+
+  const teamMembers = [
```
jpvajda, line 22: It be cool if all the team member social icons were positioned in the same location regardless of the <p> size. It looks a little odd to see these icons positioned in different locations on the page. 

Screen Shot 2021-01-27 at 9 12 43 AM

src/pages/relicans.js
caylahamann, line 335: Since we have dark mode treatment on our website, we use global css variables to allow for the easy changing between colors. What sort of color background were you going for?

caylahamann: Because right now, in switching to darkmode, this just becomes a white bar on the screen

aaronbassett: Gah! I meant to remove that inline style and put it in the scss file. It also doesn't require a background at all. I'll make that change.

jpvajda (changes requested):

src/pages/relicans.module.scss
```diff
@@ -0,0 +1,77 @@
...
+      max-width: 150px;
+      height: auto;
+      border-radius: 50%;
+  }
+  
+  .point {
+    text-align: center;
+    max-width: 400px;
+    min-height: 575px;
+    margin: auto;
```
caylahamann, line 56: I would get rid of this `margin: auto;` because it's currently causing the tiles to be uneven. (Look at Jonan Scheffler's tile) I think you need to use something like `align-items` in order to center them on a mobile view.

caylahamann: Screen Shot 2021-01-27 at 11 19 35 AM

src/pages/relicans.js
aaronbassett: It's difficult with the Bios all being such different lengths. Maybe if I moved them before the bio?

jpvajda (approved):

PR #1073, opened 2021-01-27
Relican page fixes

caylahamann:
* Fixes a ton of linting errors 
* Provides the correct class to the iframe so that it wraps with the page


src/pages/relicans.js
```diff
@@ -4,344 +4,115 @@ import SEO from '../components/Seo';
...
-          url: "https://www.linkedin.com/in/rachael-wright-chaelcodes/"
-        }
-      ]
-    }
-  ]
-
-  const teamMemberPanels = teamMembers.map(teamMember => {
-    const socialItems = teamMember.socials.map(socialNetwork => {
+  const teamMemberPanels = teamMembers.map((teamMember, i) => {
+    const socialItems = teamMember.socials.map((socialNetwork, i) => {
```
jerelmiller, line 12: It looks like you're using `i` twice, both for the outer loop and the inner one. Does that cause any problems? Would it make sense to update one of them to `j`?

src/pages/relicans.js
```diff
@@ -4,344 +4,115 @@ import SEO from '../components/Seo';
...
+                max-width: 560px;
+                iframe {
+                  width: 100%;
+                  height: 315px;
+                }
+              `}
+              width="560"
+              title="Relicans Page Video"
+              height="315"
+              src="https://www.youtube.com/embed/OlW3g631ero"
```
jerelmiller, line 107: It looks like this is the same URL as the `Video` component generates: https://github.com/newrelic/gatsby-theme-newrelic/blob/develop/packages/gatsby-theme-newrelic/src/components/Video.js#L8

By all means, that `id` should work. What error were you seeing with the `Video` component when using it? I'd much prefer that if we can since it has all of the styles nicely contained in it.

src/data/relicansData.js
```diff
@@ -0,0 +1,379 @@
+import React from 'react';
```
jerelmiller, line 1: It doesn't look like you're rendering any react components here. Do you need this import?

jerelmiller (commented):
Couple minor things for you.

src/pages/relicans.js
caylahamann: that does make more sense :) I will fix that, I was just getting some linting errors for not having a key so my immediate reaction is to add an i with a key

src/pages/relicans.js
caylahamann: @jerelmiller I totally agree. When I was doing it, the video would not appear. I'm not sure why because I was getting no errors

src/data/relicansData.js
caylahamann: It required it because of the <p> tags in the bios!

jerelmiller: I'm so dumb 😅 . How did I not see that?!?!!

src/pages/relicans.js
jerelmiller: The only thing I can think of is that for some reason the size of the video was `0x0`, so even if it was being rendered, it wasn't seen because it had no width/height. That would be my guess for why that was happening. We can look into it later. For now this is fine, though my guess is this doesn't scale with mobile well.

jerelmiller (approved):

PR #1078, opened 2021-01-28
Generate component documentation from GraphQL instead of runtime

jerelmiller:
## Description

Generates the New Relic One SDK documentation from the data in GraphQL rather than at runtime. This should help out our SEO tremendously for those pages. This PR also auto generates pages for each component/api in the SDK so that we no longer have to manually create MDX files for each component.

## Screenshots

**Before**

developer newrelic com_components_button

**After**

localhost_8000_components_button


src/components/PropList.js
```diff
@@ -179,4 +256,211 @@ PropList.propTypes = {
   ),
 };
 
+export const query = graphql`
+  fragment PropList_propTypes on NewRelicSdkPropTypeDefinition {
```
jerelmiller, line 260: Don't ask. This query was so hard 😂

zstix (approved):
🎉 This looks fantastic! Thank you so much for digging into this!

src/components/PropList.js
zstix: I remember trying to write prop types for the prop prop types and getting so turned around 😵

jerelmiller:
You're welcome! This was fun :)

PR #1086, opened 2021-01-29
Adds both the validate PR action and caching to the dev site

caylahamann:
closes https://github.com/newrelic/docs-website/issues/699

### DIFF
--- a/.github/workflows/build-cache.yml
+++ b/.github/workflows/build-cache.yml
@@ -5,10 +5,6 @@ on:
     branches:
       - develop
 
-env:
-  NODE_OPTIONS: '--max-old-space-size=4096'
-  CHOKIDAR_USEPOLLING: 1
-
 jobs:
   build:
     name: Gatsby Build
@@ -38,7 +34,7 @@ jobs:
           key: ${{ runner.os }}-gatsby-build-${{ github.run_id }}
           restore-keys: |
             ${{ runner.os }}-gatsby-build-
-            
+
       - name: Install dependencies
         if: steps.yarn-cache.outputs.cache-hit != 'true'
         run: yarn install --frozen-lockfile
--- a/.github/workflows/build-cache.yml
+++ b/.github/workflows/build-cache.yml
@@ -38,6 +38,7 @@ jobs:
           key: ${{ runner.os }}-gatsby-build-${{ github.run_id }}
           restore-keys: |
             ${{ runner.os }}-gatsby-build-
+            
       - name: Install dependencies
         if: steps.yarn-cache.outputs.cache-hit != 'true'
         run: yarn install --frozen-lockfile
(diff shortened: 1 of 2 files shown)


.github/workflows/build-cache.yml
```diff
@@ -0,0 +1,49 @@
+name: Build and cache site
+
+on:
+  push:
+    branches:
+      - develop
+
+env:
+  NODE_OPTIONS: '--max-old-space-size=4096'
+  CHOKIDAR_USEPOLLING: 1
```
jerelmiller, line 10: We can remove this for now if you'd like. We needed this on the docs site for the sheer amount of content there. It was running out of memory on builds. The dev site is much smaller and shouldn't suffer that problem for quite a while.

.github/workflows/build-cache.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Cache Gatsby build
+        uses: actions/cache@v2
+        with:
+          path: |
+            public
+            .cache
+          key: ${{ runner.os }}-gatsby-build-${{ github.run_id }}
+          restore-keys: |
+            ${{ runner.os }}-gatsby-build-
+      - name: Install dependencies
```
jerelmiller, line 41:
```suggestion
            ${{ runner.os }}-gatsby-build-
            
      - name: Install dependencies
```

.github/workflows/validate-pr.yml
```diff
@@ -0,0 +1,77 @@
...
+        if: steps.yarn-cache.outputs.cache-hit != 'true'
+        run: yarn install --frozen-lockfile
+
+      - name: Gatsby build
+        run: yarn build:production --log-pages
+        env:
+          GATSBY_EXPERIMENTAL_PAGE_BUILD_ON_DATA_CHANGES: true
+          CI: true
+
+  lint:
```
jerelmiller, line 53: Can we also run tests as part of PR validation?

jerelmiller (commented):
2 very minor things. Thanks so much for getting this in! We've needed this for so long.

.github/workflows/validate-pr.yml
caylahamann: a bunch of the tests are failing (which is why I excluded it) because I didn't know if we wanted to address that first. For example, there's tests for our old footer

jerelmiller: Ah thats good to know!

I'd say lets still add it, just not make it a required check for now. It will be a really good reminder to fix when we open up future PRs.

caylahamann: Been added in!

.github/workflows/build-cache.yml
caylahamann: I removed it in both files!

jerelmiller (approved):

Issue #726, opened 2020-09-08
Differentiate parent and child Step components so they are easier to recognize


rhetoric101:
### Make it easier to read Markdown populated with components

Since Markdown is a left-justified format, we need all the help we can get to make it easier to read Markdown that includes components. For example, it a dense Markdown topic, I find it difficult to recognize the difference between the parent and child components for steps:

```
<Steps>
<Step>asdf</Step>
<Step>kdk</Step>
</Steps>
```

### Possible solution

Consider making the parent easier to differentiate from the child by using something like this:

```
<Procedure>
<Step>asdf</Step>
<Step>kdk</Step>
</Procedure>
```

PR #1124, opened 2021-02-10
chore: referencing theme component docs

jpvajda:
This PR removes all component documentation and references the Theme component docs.


COMPONENT_GUIDE.md
```diff
@@ -1,317 +1,4 @@
...
-title: 'Example Guide'
-template: 'GuideTemplate'
-description: 'Example guide page'
----
-```
-
-It also accepts a `<Video />` component as a child, which it will place on the left side of the description.
-
-```md
-<Intro>
```
jerelmiller, line 63: The `Intro` is only available on the dev site. We should probably keep that around.

COMPONENT_GUIDE.md
```diff
@@ -1,317 +1,4 @@
...
-## Steps
-
-The `<Steps />` is a required container for the individual `<Step />` components and will autonumber from top to bottom.
-
-### Usage
-
-The Steps component accepts `<Step/>` components as its children and will increment by the number of child components.
-
-```md
-<Steps>
```
jerelmiller, line 81: `<Step>` and `<Steps>` are only available on the docs site and not in the theme. We should proabably keep this around.

jerelmiller (changes requested):

COMPONENT_GUIDE.md
jpvajda: oh... ok. why not put them in the global theme?

COMPONENT_GUIDE.md
jpvajda: same question as above. :)

jerelmiller: I'll answer here for both 😄 

The reason is because these are patterns that are specific to the dev site and are not used by other sites. Our philosophy so far has been: Only put things in the theme that we know will be shared amongst 2 or more sites. This allows us to flush out design and UX before we just share amongst all the sites. Its expensive to make changes to the theme, and we also want to avoid breaking changes there, since it is versioned. Its much easier to make a breaking change in the site itself than in the theme.

Specifically, here are the behaviors that we may not want to pull over for each of these 2 components:

* `Intro`

This component does a couple things:

1) Enlarges the text a bit and changes its text color to make it standout among the regular body text
2) Auto-detects images/videos in the intro and automatically puts them in a right column.

Its 2) that worries me in the theme. While this makes a lot of sense for the dev site, where its very guide oriented, this pattern may or may not work for the docs site (no docs site pages uses this kind of pattern). If we were to extract this to the theme, we'd likely want to take out the behavior to do the auto columns. And if we do that, then it means we'd need to customize that component anyways if we pulled it back into the dev site.

For now, we'd rather just iterate on this in the dev site until we have a need for this behavior across all sites for their intros.

* `Step`/`Steps`

These components are very guide oriented and much better suit the patterns on the dev site instead of the docs site. We may decide we like this pattern and want to bring it to all the sites, but in doing so, I'd recommend a rename ([its currently confusing](https://github.com/newrelic/developer-website/issues/726)).

These components are also packed with all of the functionality from the recent innovation week. That work may or may not be specific to the dev site, so if we extract, we'd want to make decisions on how we can best make sure that functionality can scale across all use cases. Again, I'd like to wait to understand the use case on say, the docs site, before we fully extract this so that we make sure its API covers the cases we need without needing to make breaking changes.

Hopefully that helps :)

jpvajda: make sense! I'll add these back to this file, have we documented this _philosophy_ anywhere? this may be something useful to put into our Documentation of the theme.

COMPONENT_GUIDE.md
```diff
@@ -1,74 +1,18 @@
...
-
-```md
-<Intro>
-This is a description for the markdown guide.
-
-<Video id="zxunt1u1as" type="youtube"/>
-</Intro>
-```
-
-If there is a more than plain text and a `<Video />` (such as a code snippet or another component) the content will be posted on the left side below the description.
```
jerelmiller, line 70: It looks like this stuff is still relevant to the `Intro` component. Would you mind restoring this section? The rest of the changes look great!

COMPONENT_GUIDE.md
jerelmiller: > have we documented this philosophy anywhere

😅 ...... no

COMPONENT_GUIDE.md
jpvajda: it is done

jerelmiller (approved):

PR #1135, opened 2021-02-19
Add terraform minisite to sidenav

jaesius:
## Side nav update
I added the Terraform mini-site and its children developer guides to the side nav.


zstix (approved):
🎉

PR #1151, opened 2021-03-04
Update navigation to use theme components

jerelmiller:
## Description

Now that the theme has a bunch of "native" nav components, this PR brings them in. This helps bring much needed consistency between the dev and docs sites. In addition, this PR also:

* Moves the `NewRelicIcon` definitions to shadow the `newrelic` icon set from the theme, so that we can use the `Icon` component.
* Migrates the nav to a `yml` file to make it easier to change and to bring it inline with the docs site.

## Screenshot(s)

**Before**
Screen Shot 2021-03-03 at 8 59 39 PM

**After**

Screen Shot 2021-03-03 at 8 59 46 PM


zstix (approved):
🎉  So much cleanup! This is awesome

PR #1153, opened 2021-03-04
Correct Ali's LinkedIn URL

DavidCarrington:
Simple URL correction since the existing URL leads to an error page "This profile is not available".


jerelmiller (approved):
Good catch! Thanks for updating.

PR #1160, opened 2021-03-06
Upgrade to React 17

jerelmiller:
## Description

Upgrades the site to React 17. This is the first step to getting to Gatsby v3.


package.json
```diff
@@ -37,8 +37,8 @@
     "prism-react-renderer": "^1.1.1",
     "prismjs": "^1.21.0",
     "prop-types": "^15.7.2",
-    "react": "^16.12.0",
-    "react-dom": "^16.12.0",
+    "react": "^17.0.1",
```
caylahamann, line 40: are there any concerns with us upgrading to react 17?

caylahamann (approved):
This looks good to me! Just posed a question for my own edification. I also assume that we are first going to make these changes on the developer site before applying it the docs site as a whole?

package.json
jerelmiller: Shouldn't be! Check out the preview! https://pr-1160.dlyi50rq9kt6c.amplifyapp.com/

Its fully backwards compatible, so as far as I'm aware, nothing should break as a result of the upgrade.

PR #1164, opened 2021-03-08
chore(podcasts): Add The Relicans Launchies and Polyglot podcast embeds

aaronbassett:
## Description

Add Buzzsprout embeds for the other Relicans Podcasts

- https://twitter.com/LaunchiesShow
- https://twitter.com/PolyglotShow

## Reviewer Notes

On the [Podcasts page](https://developer.newrelic.com/podcasts) there should now be 3 embeds. Two listed above and the original Observy McObserveface

## Screenshot(s)

CleanShot 2021-03-08 at 15 39 47@2x

### DIFF
--- a/src/pages/podcasts.js
+++ b/src/pages/podcasts.js
@@ -1,11 +1,26 @@
-import React from 'react';
-import cx from 'classnames';
-import SEO from '../components/Seo';
 import PageLayout from '../components/PageLayout';
+import React from 'react';
+import SEO from '../components/Seo';
+import cx from 'classnames';
 import podcastsHeader from '../images/podcasts/podcasts.jpg';
 import styles from './podcasts.module.scss';
 
 const PodcastsPage = () => {
+  const podcastIds = ['1225223', '1677727', '1677670'];
+
+  const podcastEmbeds = podcastIds.map((podcastId) => {
+    return (
+      <section className={cx(styles.section, styles.player)} key={podcastId}>
+        <div>
+          <iframe
+            title="buzzsprout"
+            src={`https://www.buzzsprout.com/${podcastId}?client_source=large_player&iframe=true&referrer=https://www.buzzsprout.com/${podcastId}.js?container_id=buzzsprout-large-player-${podcastId}&player=large`}
+          />
+        </div>
+      </section>
+    );
+  });
+
   return (
     <>
       <SEO />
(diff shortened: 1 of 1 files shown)


src/pages/podcasts.js
```diff
@@ -1,11 +1,26 @@
...
 import SEO from '../components/Seo';
-import PageLayout from '../components/PageLayout';
+import cx from 'classnames';
 import podcastsHeader from '../images/podcasts/podcasts.jpg';
 import styles from './podcasts.module.scss';
 
 const PodcastsPage = () => {
+  const podcastIds = ['1225223', '1677727', '1677670'];
+
+  const podcastEmbeds = podcastIds.map((podcastId) => {
```
jerelmiller, line 11: [nit] I generally prefer to have all the JSX inline so that you don't have to bounce back and forth when understanding this part of the UI.

This change is GREAT by the way. I really like that this makes it easy to add new podcast sections in the future.

src/pages/podcasts.js
```diff
@@ -1,11 +1,26 @@
...
+    return (
+      <section className={cx(styles.section, styles.player)} key={podcastId}>
+        <div>
```
jerelmiller, line 14: [nit] It looks like this used to have an `id` (which looked useless, so thanks so much for removing that!). Now that this is empty, do we need this wrapping div? Could we instead just put the `iframe` here?

src/pages/podcasts.js
```diff
@@ -1,11 +1,26 @@
...
+          <iframe
+            title="buzzsprout"
```
jerelmiller, line 16: Not sure if this matters or not, but now that we have 3 podcast sections, these all have the same `title` attribute. Should this reflect what the podcast is?

jerelmiller (commented):
Hey @aaronbassett!  Thanks so much for this contribution! Awesome that we have so much happening at New Relic that we can show off.

I've added a few `[nit]` comment, which you can accept or reject. Feel free to use your best judgement on whether you'd like to address this or not!

aaronbassett:
Thanks for the review @jerelmiller, good catches! I've made the requested changes

jerelmiller (approved):
Thanks so much for addressing those comments! This PR looks 🔥 🔥 🔥 . Excited to get this out!

PR #1172, opened 2021-03-11
chore(podcasts): remove mentions of only having one podcast

AishaBlake:
## Description

We recently added Polyglot and Launchies podcasts to the site but didn't change the text that comes _before_ the podcast player embeds. Three paragraphs of the existing page content refer to only having a single podcast and therefore no longer make sense. I caught this as we were ending the stream where we made this change and would have forgotten to go back and fix it if not for @xomiamoore!

## Reviewer Notes

No code changes, but please do read the whole page and confirm it still makes sense?

## Related PR

* #1164

### DIFF
--- a/src/pages/podcasts.js
+++ b/src/pages/podcasts.js
@@ -35,9 +35,9 @@ const PodcastsPage = () => {
                 perfect for each other.
               </p>
               <p>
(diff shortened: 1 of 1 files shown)


jerelmiller (approved):
Nice! Thanks so much for updating this 😄

Issue #1163, opened 2021-03-08
Table SDK component documentation shows an error


zstix:
## Description
```
Invariant Violation: Unable to find node on an unmounted component.
```

### Doc information (don't delete this section)
* https://developer.newrelic.com/components/table
* **OS**: Mac OS 11.2.2
* **Browser**: Chrome 88.0.4324.192
* **Device**: Unknown

<img width="1666" alt="Screen Shot 2021-03-11 at 7 48 52 AM" src="https://user-images.githubusercontent.com/1395158/110805384-43158880-823e-11eb-8319-74e900378609.png">

### Acceptance Criteria

- [ ] This has been fixed by One Core, so we need to pull in a new version of the SDK bundle to fix this problem.
- [ ] Document where you can obtain the release verison in our bundle [update docs section](https://github.com/newrelic/developer-website/blob/develop/CONTRIBUTING.md#updating-the-documentation-bundle)

jpvajda:
I wonder if this has always been there since our last SDK update? Perhaps it's something the Nr1 Core Team can help us with.

jpvajda:
I'll see if the NR1 Core Hero can give us some direction here.

jpvajda:
Still waiting on a response... 😓

jpvajda:
I believe our dev site should render these example tables, as it's rendered on NR1 core site. That seems to be what the error is indicating.

![Screen Shot 2021-03-15 at 5 58 41 PM](https://user-images.githubusercontent.com/1395158/111236418-6df23a80-85b8-11eb-9d8f-7154894c242f.png)

![Screen Shot 2021-03-15 at 5 58 47 PM](https://user-images.githubusercontent.com/1395158/111236421-6f236780-85b8-11eb-9394-14689948c84a.png)

jpvajda:
Interesting this does render locally... 🤔  

![Screen Shot 2021-03-15 at 6 07 26 PM](https://user-images.githubusercontent.com/1395158/111236860-739c5000-85b9-11eb-8f15-dcaa1b45f3e7.png)

LizBaker:
this is broken locally for me. it looks to be a react versioning issue, if I roll the react version we're using back to 16.8 the tables load just fine. the dev site upgraded to react 17 12 days ago so I'm thinking that's when we ran into this issue. I would think it was a version mismatch/issue with react-live which we're using to render these previews, but that appears to function normally for everything but tables, continuing to poke around

jpvajda:
[...]

PR #1196, opened 2021-03-25
Fix conflicting react versions/update SDK

LizBaker:
## Description

This PR updates the version of (and way we add) react on the window and updates the SDK

## Reviewer Notes

To test this, make sure the live preview is working for both the table components and _**all**_ the charts
Additionally, poke around some other components/pages to be sure all is well

Closes https://github.com/newrelic/developer-website/issues/1163


gatsby-config.js
```diff
@@ -1,6 +1,9 @@
 const quote = (str) => `"${str}"`;
 
 module.exports = {
+  flags: {
```
jpvajda, line 4: Is this just a change we are including in this PR that is seperate from the SDK issue? or is this required to fix it? I was just curious.

jpvajda (approved):
LGTM , just has a question on the inclusion of the `LAZY_IMAGES = true`

jpvajda (changes requested):
Actually I tested on the amplify URL and this page isn't loading....

go to: https://pr-1196.dlyi50rq9kt6c.amplifyapp.com/components/table and see the failure....

Screen Shot 2021-04-22 at 3 51 02 PM

gatsby-config.js
LizBaker: oh, yeah that was just a suggestion to add by Jerel but I can definitely remove it

jerelmiller: Ya it doesn't seem to work anyways, I think because this branch is using an old version of `gatsby-image` that doesn't support lazy images. With the upgrade to v3, I think this flag is obsolete. Feel free to remove.

jerelmiller:
@LizBaker I think @jpvajda is experiencing the race condition we chatted about. I see mixed results on the component pages when loading that page. Sometimes it loads, sometimes it doesn't, depending on whether the SDK bundle has finished processing in time.

We may need a mechanism to detect when the bundle is fully loaded so that the component documentation can put itself in a loading state until it is. I think right now those pages assume the SDK is already loaded.

jpvajda:
I'm also not sure if amplify previews are a good representation of how this will actually work once deployed, I checked the functionality on Chrome and it was fine....

jerelmiller:
> I'm also not sure if amplify previews are a good representation of how this will actually work once deployed

On the contrary, these are actually the _best_ representation of how they will work when deployed since it builds through the same environment (amplify). There are only minor differences between the preview environment and the production environment:

* The URL (which is fairly obvious 😄 )
* The NR environment that defines where our instrumentation goes (we label preview builds as a "staging" environment)
* The environment is destroyed when the PR is merged

The rest of it mimics production exactly, which also includes amplify configuration, such as the Amplify redirects configuration.

If there is an issue you see in the amplify preview, its 99% guaranteed to happen in prod as well (the only exception I can think of would be some code that may rely on the domain name in the browser).

Hope this helps! 😄

jpvajda:
@jerelmiller thanks for the explaination! If that is true this is also not rendering in Safari 😓

jerelmiller:
Good to know! I _think_ this is actually a race condition between the SDK bundle loading and when the component preview tries to render. _If_ that is the case, then this should affect all browsers. I chatted with @LizBaker a bit on some ideas to try and fix this issue.

jerelmiller:
> this is actually a race condition 

To clarify, this would explain why sometimes it loads just fine, and other times it blows up

LizBaker:
added a possible solution. this way we can track the load of the sdk and error handle. definitely open to feedback about the way its being done and _where_ its being done 💯

zstix (approved):
Looks good to me :tada:

PR #1223, opened 2021-04-02
chore: adds tessen configuration

jpvajda:
This PR adds the configuration for tessen and remove the individual GA tracking code so we can leverage all the goodness of Tessen and Segment. 🎸

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -22,7 +22,7 @@ module.exports = {
           component: require.resolve('./src/layouts'),
           mobileBreakpoint: '760px',
         },
-        prism: {
+       prism: {
           languages: ['yaml', 'sass', 'scss', 'java'],
         },
         splitio: {
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -16,14 +16,13 @@ module.exports = {
     {
       resolve: '@newrelic/gatsby-theme-newrelic',
       options: {
-        gaTrackingId: 'UA-3047412-33',
         layout: {
           contentPadding: '2rem',
           maxWidth: '1700px',
           component: require.resolve('./src/layouts'),
           mobileBreakpoint: '760px',
         },
-       prism: {
+        prism: {
           languages: ['yaml', 'sass', 'scss', 'java'],
         },
         splitio: {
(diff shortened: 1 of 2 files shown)


LizBaker:
if you're removing ga you can also remove the `onRouteUpdate` function from gatsby-browser.js to keep things clean

jpvajda:
@LizBaker Ok, I see on the Docs Site we handle this a little different for `OnRouteUpdate` Should I take that approach on the developer site in stead of removing it all together? 

https://github.com/newrelic/docs-website/blob/develop/gatsby/on-route-update.js

```
const onRouteUpdate = ({ location, prevLocation }) => {
  if (
    window.newrelic &&
    location.hash &&
    location.pathname !== prevLocation?.pathname
  ) {
    window.newrelic.addPageAction('hash_request', { hash: location.hash });
  }
};
```

LizBaker:
@jpvajda I don't think so. I'm not sure why we're tracking that but since it doesn't have to do with tessen or GA (looks like it's the browser agent) I'd ignore it for this PR

gatsby-config.js
```diff
@@ -16,14 +16,13 @@ module.exports = {
...
       options: {
-        gaTrackingId: 'UA-3047412-33',
         layout: {
           contentPadding: '2rem',
           maxWidth: '1700px',
           component: require.resolve('./src/layouts'),
           mobileBreakpoint: '760px',
         },
-       prism: {
+        prism: {
```
zstix, line 25:
```suggestion
       prism: {
```

gatsby-config.js
```diff
@@ -16,14 +16,13 @@ module.exports = {
     {
       resolve: '@newrelic/gatsby-theme-newrelic',
       options: {
-        gaTrackingId: 'UA-3047412-33',
```
zstix, line 19: Are we positive we're not using this tracking ID for anything else?

Just want to make sure that the addition of Tessen (and Segment) replaces everything we were getting from Google Analytics.

zstix (commented):
I had one question and a minor nitpick change. If we're confident that this the path forward, I'm good with it!

gatsby-config.js
jpvajda: according to Zoe we no longer need this in the site config. I looked at the Docs site and how it was [implemented](https://github.com/newrelic/docs-website/wiki/Event-Data--and--Instrumentation)  and the way I understand it to work is: we rely on this configuration with tessen to pass all the data we need to segment and GA, and within Segment we have the GA tracking ID  configured. 

``` tessen: {
          product: 'DOC',
          subproduct: 'TDOC',
          segmentWriteKey: 'AEfP8c1VSuFxhMdk3jYFQrYQV9sHbUXx',
          trackPageViews: true,
          pageView: {
            name: 'pageView',
            category: 'DocPageView',
            getProperties: ({ location, env }) => ({
              path: location.pathname,
              env: env === 'production' ? 'prod' : env,
            }),
          },
        },

```

LizBaker (approved):

jpvajda:
Once I test and confirm this is working with the OSS SIte, I'll cut a release for the developer site.

Issue #1235, opened 2021-04-08
Add `NODE_EXTRA_CA_CERTS` docs


alexronquillo:
### Summary

We need to write content to help debug a situation where the user is interacting with `nr1` behind a proxy.

The solution:

> Add their own corporate certificate by NODE_EXTRA_CA_CERTS's environment variable.

### What type of documentation change are you suggesting?

Update to existing docs

### Motivation

This will help users debug a specific proxy-related issue.

### Writing the documentation

If you are interested in helping us create this documentation please let us know. 
We encourage community contributions and can more quickly move your idea to publication if you suggest changes, or create an outline or rough draft. We'll collaborate with you to hone whatever content you provide.

jpvajda:
There is some documentation on proxy setups here: not sure if this is relevant to this case: https://developer.newrelic.com/build-apps/build-hello-world-app#proxy-browser-setup

PR #1246, opened 2021-04-12
chore: Add advanced config guide

alexronquillo:
## Description

Adds advanced config guide to establish a source-of-truth guide for this content, rather than it living in the "Hello, world" tutorial.

Fixes: https://github.com/newrelic/developer-website/issues/1235

### DIFF
--- a/src/markdown-pages/build-apps/build-hello-world-app.mdx
+++ b/src/markdown-pages/build-apps/build-hello-world-app.mdx
@@ -275,6 +275,20 @@ Select the accounts you want to subscribe to the application, and then click **U
 
 When you return to the Apps page, you'll see the launcher for your new application.
 
+## Proxy browser setup
+
+Generally, you don't need to know how your application code is loaded in New Relic One. But understanding how this works can help you if you encounter issues (for example, with a proxy or browser setup).
+
+When developing, your code is served from your local laptop by using an nr-local.net subdomain. This domain points to 127.0.0.1, and will use a variety of ports to connect, including 9973.
+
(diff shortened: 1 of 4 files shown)


src/markdown-pages/build-apps/build-hello-world-app.mdx
```diff
@@ -275,26 +275,6 @@ Select the accounts you want to subscribe to the application, and then click **U
 
 When you return to the Apps page, you'll see the launcher for your new application.
 
-## Advance configurations
-
-### Extend webpack configuration
-
-You can use our `.extended-webpackrc.js` file to extend the webpack 4 configuration. This allows you to add your own aliases, scripts, and configurations when locally serving or deploying your app.
-
-### Proxy browser setup
```
jpvajda, line 284: Is this moving to another location? It was something Miguel felt was important to bring over from the AEM site, so I actually had to go find it and add it to this doc. 😄

jpvajda (changes requested):
Just needed to confirm my question on the Proxy Browser setup  content

src/markdown-pages/build-apps/build-hello-world-app.mdx
alexronquillo: Yes. I don't think this makes sense to have in "Hello, World", so it's moving. But that's in progress. That change should be in by this week 🤞 If you think it needs to stay in Hello, World until the change is in, I can add it back for now.

alexronquillo: Originally this change was part of that bigger restructure branch, so I overlooked that it got removed when I moved this to its own branch.

alexronquillo: @jpvajda I just added it back for now, so that we can get this through.

jpvajda (approved):

PR #1282, opened 2021-04-26
fix: referrer field not being sent on sign up

aswanson-nr:
## Description
The `gatsby-plugin-newrelic` dependency was not properly passing the referral field to segment. This updates to the newest version with a fix.

Partly closes https://github.com/newrelic/docs-website/issues/1816


roadlittledawn (approved):
:shipit:

PR #1299, opened 2021-05-04
Force trailing slashes on page paths

### LINKED ISSUE #1296: Only 15 pages indexed in Swiftype
## Description

Only 15 pages on the developer site are indexed in Swiftype. I believe there should be... 100+ pages indexed? 

## Screenshots
Screen Shot 2021-05-03 at 16 04 01

roadlittledawn:
## Description

Adds theme config option to force trailing slashes to try to make sure the sitemap includes trailing slashes to hopefully allow swiftype to index without encountering 302s.

- I _think_ this will close #1296

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -82,7 +82,9 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
     }
 
     createPage({
-      path: frontmatter.path || slug,
+      path: frontmatter.path
+        ? path.join(frontmatter.path, '/')
+        : path.join(slug, '/'),
       component: path.resolve(`src/templates/${frontmatter.template}.js`),
       context: {
         slug,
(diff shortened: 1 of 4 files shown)


gatsby-node.js
```diff
@@ -82,7 +82,9 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
     }
 
     createPage({
-      path: frontmatter.path || slug,
+      path: frontmatter.path
+        ? path.join(frontmatter.path, '/')
+        : path.join(slug, '/'),
```
aswanson-nr, line 87: Are we sure that a `slug` won't already have a trailing slash?

aswanson-nr (approved):
One comment! Please ignore it if you already thought of it.

gatsby-node.js
roadlittledawn: i'm pretty sure it won't.

for this line, this is for the `allMdx` nodes, which pulls `slug` off of `fields`. we add that node field for all nodes with type `Mdx` in `onCreateNode` lifecycle in the same file on [line 142](https://github.com/newrelic/developer-website/blob/614ba41a3d133b3728ef299dcf26e574c7741c21/gatsby-node.js#L142). in `createFilePath()` we set the `slug` value with the option `trailingSlash: false`.

PR #1316, opened 2021-05-13
chore: Clarify what Nerdpacks basic users can use

alexronquillo:
## Description

Based on conversations, it's not true that basic users can use any non-FSO Nerdpacks. They can only use Nerdpacks that New Relic specifically permits basic users to use.


jaesius (approved):

zuluecho9:
Looks good, Alex. Thank you for trying to deal with this confusing area.

PR #1334, opened 2021-05-24
fix: update to theme version that handles blocked cookies

aswanson-nr:
# Summary
Handle case when cookies are blocked.

Closes https://github.com/newrelic/docs-website/issues/2293

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "@emotion/styled": "^11.3.0",
     "@mdx-js/mdx": "^1.6.19",
     "@mdx-js/react": "^1.6.19",
-    "@newrelic/gatsby-theme-newrelic": "^2.4.0",
+    "@newrelic/gatsby-theme-newrelic": "^2.4.2",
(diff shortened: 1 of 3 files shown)


roadlittledawn:
so far seems to be working okay. but i did see that embed pages seem to use dark mode when they should default to light mode. example: http://localhost:8000/automate-workflows/get-started-terraform/embed

roadlittledawn:
Search works for me. And I don't think developer has any Collapser components

aswanson-nr:
> so far seems to be working okay. but i did see that embed pages seem to use dark mode when they should default to light mode. example: http://localhost:8000/automate-workflows/get-started-terraform/embed
> 
> 

Here's the issue I put together for that problem https://github.com/newrelic/developer-website/issues/1335.

LizBaker (approved):
i'm good with moving forward despite the embed issue since we have it documented and someone having cookies entirely blocked is less likely

PR #1347, opened 2021-05-27
Update my bio

aaronbassett:
## Description

My bio information was outdated


rudouglas (approved):
LGTM the ESLint issue could be changed but it's not a blocker by any means :D

PR #1355, opened 2021-05-28
chore: add o11y pack fake data

### LINKED ISSUE #1331: [Fake API Data] Create sample data set
## Summary

We want to create some sample JSON we can then load into Gatsby GraphQL and build/run time. Depending on what shape we get data from the final API, we may still create a script that fetches the data and ultimately transforms it to a format that best suits our needs.

### Shape of data
We could either create one JSON file with all the data we need for each pack OR we could create multiple JSON files (especially if we have multiple data sources) that we stitch together when loading into GraphQL (similar to how opensource does it).

Perhaps something like this may make sense, where each `.json` file contains all the data we would need for now.

```
│  data
│  ├── observability-packs
│  │   ├── browser
│  │   │   ├── browser.json
│  │   │   └── images
│  │   └── mysql
│  │       ├── images
│  │       └── mysql.json
```

### Data sources
We'll source data from the main observability pack GraphQL API for most pack data.

We'll also need to source data from github re: some pack info. Examples: contributors, copy/description, images. 

## Acceptance criteria
- [x] Confirm with @kidk when the real API will be completed.
- [x] We have JSON file(s) that approximate what we will have access to
- [x] The JSON file(s) have sufficient data that's conducive to the needs of our pages / templates

## Helpful links
- https://hackmd.io/6rTdhgu5TqacJ2V3yMURWA?both#GraphQL-Schema
- Related Resources data (developer, docs)
- [Opensource github data](https://github.com/newrelic/opensource-website/tree/develop/src/data)

aswanson-nr:
## Description

Some fake data to get us started. This is grabbed from the future graphql endpoints and modified to include alerts for some of the packs. I also removed all dashboards and alerts from one of the packs to help with testing, since they are not required.

Closes #1331

### DIFF
--- a/src/data/observability-packs.json
+++ b/src/data/observability-packs.json
@@ -158,7 +158,7 @@
       "description": "Official New Relic pack for Consul\nUse this pack together with the New Relic Consul On Host Integration to get insight into the performance of your Consul instances.\n\n",
       "icon": "https://raw.githubusercontent.com/newrelic/newrelic-observability-packs/v0.3.1/packs/consul/icon.png",
       "id": "Q29uc3Vs",
-      "level": "NEWRELIC",
+      "level": "VERIFIED",
       "logo": "https://raw.githubusercontent.com/newrelic/newrelic-observability-packs/v0.3.1/packs/consul/logo.png",
       "name": "Consul",
       "website": "https://www.consul.io/"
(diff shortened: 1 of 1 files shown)


src/data/observability-packs.json
```diff
@@ -0,0 +1,535 @@
...
+        {
+          "name": "Throughput",
+          "definition": "",
+          "url": ""
+        }
+      ],
+      "description": "Official New Relic pack for Apache\nUse this pack together with the New Relic Apache On Host Integration to get insight into the performance of your Apache instances.\n\n",
+      "icon": "https://raw.githubusercontent.com/newrelic/newrelic-observability-packs/v0.3.1/packs/apache/icon.png",
+      "id": "QXBhY2hl",
+      "level": "NEWRELIC",
```
LizBaker, line 35: should we change the level on a couple of these? maybe add a `COMMUNITY` or two (not actually sure of the official terms)

aswanson-nr: Good call! I'll add a few of the other levels

LizBaker (approved):

PR #1359, opened 2021-06-02
Ruairi/install component

rudouglas:
This includes an Install button and Dropdown button with list of options. These are just samples for now to showcase functionality

### DIFF
--- a/src/components/InstallButton.js
+++ b/src/components/InstallButton.js
@@ -5,7 +5,7 @@ import { css } from '@emotion/react';
 
 const sampleItems = new Array(10).fill().map((_, i) => i + 1);
 
-const createMenuItems = (children, items) => {
+const createMenuItems = (items) => items.map((i) => <Dropdown.MenuItem>{i}</Dropdown.MenuItem>)
   const menuItems = [];
   for (const item of items) {
     menuItems.push(<Dropdown.MenuItem>{item}</Dropdown.MenuItem>);
--- a/src/components/InstallButton.js
+++ b/src/components/InstallButton.js
@@ -37,7 +37,7 @@ const InstallButton = ({ children, title, ...props }) => {
           margin-right: 2px;
         `}
       >
-        Install
+        {title}
       </Button>
       <Dropdown>
         <Dropdown.Toggle
--- a/src/components/InstallButton.js
+++ b/src/components/InstallButton.js
@@ -1,0 +1,63 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Button, Dropdown, Link } from '@newrelic/gatsby-theme-newrelic';
+import { css } from '@emotion/react';
+
+const sampleItems = new Array(10).fill().map((_, i) => i + 1);
+
+const createMenuItems = (children, items) => {
+  const menuItems = [];
+  for (const item of items) {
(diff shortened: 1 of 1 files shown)


src/components/InstallButton.js
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Button, Dropdown, Link } from '@newrelic/gatsby-theme-newrelic';
+import { css } from '@emotion/react';
+
+const sampleItems = new Array(10).fill().map((_, i) => i + 1);
+
+const createMenuItems = (children, items) => {
```
aswanson-nr, line 8: This could be simplified to 
```suggestion
const createMenuItems = (items) => items.map((i) => <Dropdown.MenuItem>{i}</Dropdown.MenuItem>)
```
I omitted `children` since it doesn't appear to be used.

src/components/InstallButton.js
```diff
@@ -0,0 +1,63 @@
...
+  return `https://one.newrelic.com/launcher/nr1-core.explorer`;
+};
+
+const InstallButton = ({ children, title, ...props }) => {
+  return (
+    <div
+      css={css`
+        display: flex;
+        flex-direction: row;
+        float: right;
```
aswanson-nr, line 26: `float` is traditionally used on images or other content that you need the rest of the pages content to flow around. So I'm a little wary about using this on a button. What is it accomplishing here? I'm sure there's another way to get the desired behavior.

aswanson-nr (changes requested):

src/components/InstallButton.js
rudouglas: @aswanson-nr I only kept that in because we will be using `children` eventually but I can take it out for now 🚸

src/components/InstallButton.js
rudouglas: @aswanson-nr ooh yes good call I was just using that for display purposes for screenshots, probably don't need to include any sort of styling like that in the component itself do we? Leave that in the layout?

aswanson-nr: Leaving it to the parent component/layout sounds like the way to go! That way we can use it in multiple contexts, if needed.

src/components/InstallButton.js
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Button, Dropdown, Link } from '@newrelic/gatsby-theme-newrelic';
+import { css } from '@emotion/react';
+
+const sampleItems = new Array(10).fill().map((_, i) => i + 1);
+
+const createMenuItems = (items) => items.map((i) => <Dropdown.MenuItem>{i}</Dropdown.MenuItem>)
+  const menuItems = [];
```
aswanson-nr, line 9: Looks like those auto changes screwed up a bit here 🙃

rudouglas: whoopsy

rudouglas: i'll take another look at this not sure what's going on

rudouglas: 🤷

rudouglas: aha ok it kept part of both functions for some reason, should be fixed now 😀

PR #1361, opened 2021-06-02
Add o11y pack details template

LizBaker:
WIP


gatsby-node.js
```diff
@@ -1,8 +1,7 @@
 const path = require(`path`);
 const { execSync } = require('child_process');
 const { createFilePath } = require('gatsby-source-filesystem');
-
-const MAX_RESULTS = 5;
+const observabilityPacks = require('./src/data/observability-packs.json');
```
roadlittledawn, line 4: These should already be in Gatsby's graphql in `allObservabilityPacks` type

gatsby-node.js
```diff
@@ -97,6 +96,22 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
     });
   });
 
+  observabilityPacks.forEach((pack) => {
+    const { name } = pack;
+    const urlNormalizedName = kebabCase(name);
```
roadlittledawn, line 101: not super familiar with `kebabCase` but we noticed yesterday there may be special characters in names, so we probably want to slugify the strings to be safe. not sure if `kebabCase` does handles that already.  i think we have a slugify function in docs repo somewhere.

roadlittledawn:
i _think_ we may also want to add a node field for pack slugs (which are used to create the page paths) on the `allObservabilityPacks` type via `gatsby-node.js`. that way we have a reliable to read that from the node data and create the link (like on the landing page)

gatsby-node.js
roadlittledawn: found it! here's our [slugify util in docs repo](https://github.com/newrelic/docs-website/blob/develop/scripts/utils/slugify.js)

LizBaker: yes! i actually wrote my own and then noticed we had this and wasn't sure how weird it was to have extra characters in the slug. we could filter out anything but alphanumeric and hyphens 🤷‍♀️ or we could just use the ID in the slug and not the title

gatsby-node.js
LizBaker: good point! i started before that was hooked up

nr-kkenney (approved):

PR #1363, opened 2021-06-03
Ruairi/search functionality

rudouglas:
Uses the filter bar and react `useState()` to filter by search term on both Pack name and description. There is no `dependency` data in the packs at the moment, I think filtering on this over description would be best
Screenshot 2021-06-03 at 15 44 15
Screenshot 2021-06-03 at 15 44 25

### DIFF
--- a/src/pages/observability-packs.js
+++ b/src/pages/observability-packs.js
@@ -12,7 +12,6 @@ const ObservabilityPacksPage = ({ data, location }) => {
     allObservabilityPacks: { nodes: o11yPacks },
   } = data;
   const [filteredPacks, setFilteredPacks] = useState(o11yPacks);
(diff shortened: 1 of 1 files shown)


rudouglas:
Updated this to include Filtering and Sorting
Refactored the Search code so that the filter/sort/search all work together, and each user choice is maintained.

E.g. when you search for `Co` then sort/filter only packs containing `Co` will be included


https://user-images.githubusercontent.com/17122543/120707074-c64b2100-c4b1-11eb-877e-b2f3d541b171.mov

roadlittledawn:
Looking great! I gave it a test spin and worked well too. Good call on matching not only on name but the description if no match on name. 🤘

PR #1370, opened 2021-06-04
Ruairi/search query params

rudouglas:
Get and Set Query Params in the URL including Search/Filter/Sort


jpvajda (approved):
code looks good and the functionality checks out.

PR #1371, opened 2021-06-04
feat: pack list view

rudouglas:
Added alternate List view for all packs which is responsive to view size

Changed the name of a couple of components to keep them in line with functionality:

`PackList` -> `PackGrid` (the list view uses `PackList` and vice-versa
`PackTile` -> `PackGridTile` (to differentiate from `PackListTile`

### DIFF
--- a/src/components/PackListTile.js
+++ b/src/components/PackListTile.js
@@ -2,6 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/react';
 import { Link, Icon, Surface } from '@newrelic/gatsby-theme-newrelic';
+import NewRelicIcon from '@newrelic/gatsby-theme-newrelic/src/icons/logo/newrelic.js';
+
 
 const PackListTile = ({
   name,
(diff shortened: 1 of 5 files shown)


src/components/PackListTile.js
```diff
@@ -0,0 +1,134 @@
...
+  <Surface
+    as={Link}
+    to={to}
+    className={className}
+    base={Surface.BASE.PRIMARY}
+    interactive
+    css={css`
+      display: grid;
+      border-radius: 0.25rem;
+      position: relative;
```
aswanson-nr, line 24: I don't think we need this `position: relative` here, it can have some interesting interactions with Grid that I don't think we want https://blogs.igalia.com/mrego/2016/05/27/css-grid-layout-and-positioned-items/

aswanson-nr: It won't let me comment on the grid version, but we should also remove it there.

aswanson-nr (changes requested):
Looks good! Although I looked at it before the logo image changes, I'll review those next.

src/components/PackListTile.js
rudouglas: oooh ok cool i've removed those, and made a couple of CSS changes for dark mode in the newest commits

jpvajda:
This looks good, I think we can work on a better way to display the logos in the nit issue #1375

jpvajda (approved):

aswanson-nr (approved):

aswanson-nr:
The only thing I'm noticing is that some of the pack logos are svgs and don't render properly. @jpvajda I'm guessing that's okay for the first pass since we don't have any file type requirements on the logos currently.

jpvajda:
@aswanson-nr I'll add that to our nit list #1375

Issue #1374, opened 2021-06-04
Dev Site SDK Component Docs updates


jpvajda:
## Issues

These problems were report by `Jeremias Menichelli` In Slack, `#gpe-news`

### 1
Ignore render in components methods. We had the same issue in One Core docs. The rule of thumb is to ignore methods starting with _ as they are private and methods starting with render including render itself too as it's the method React needs, but it's not public nor should be used

<img width="1112" alt="Screen Shot 2021-06-04 at 11 15 23 AM" src="https://user-images.githubusercontent.com/1395158/120844827-3480f980-c52d-11eb-9e63-346b5112bffc.png">

### 2

Ignore empty methods definition. A function that doesn't specify any parameters and returns undefined doesn't add up to anything to the user, even when the function just doesn't take any. In this case, for the children there are plenty of examples above where the user can infer how the children prop is used

<img width="1104" alt="Screen Shot 2021-06-04 at 11 17 05 AM" src="https://user-images.githubusercontent.com/1395158/120844863-3fd42500-c52d-11eb-9905-a73ed2b366c3.png">

### 3 

If you upgrade your docs bundle you will notice we improved the typedefs parsing, so instead of {Object you will actually get a type definition from the docs

<img width="752" alt="Screen Shot 2021-06-04 at 11 18 39 AM" src="https://user-images.githubusercontent.com/1395158/120857007-e3c5cc80-c53d-11eb-8998-eed2064c7630.png">

> The comment from above relates to another improvement, once you have that typedef, you will need to:

### 4

List that typedef below so the user knows the interface of the parameter (we do that in our docs, every time we encounter a typedef in a prop we add them in a typedef map we later render at the end, check for example these type definitions at the bottom https://one-core.datanerd.us/components/nerdgraphquery/?tab=code)
The other option is to just read the typedefs structure and print it in place.

### 5 
We recently fixed the returns / return tag parsing, but you need to correctly consume it in your docs code, I am seeing a lot of undefined in the function templates. Once you got that corrected, you will get richer function definitions like this one, for example

<img width="694" alt="Screen Shot 2021-06-04 at 11 24 32 AM" src="https://user-images.githubusercontent.com/1395158/120857066-f809c980-c53d-11eb-8f85-d126c5bb47f2.png">

### Acceptance Criteria

- [x] Get all the new components pulling in
- [x] Get the Typedefs populating correctly
- [x] Ignore render in components methods
- [x] Consume return tag parsing correctly
[...]

PR #1376, opened 2021-06-04
chore: updated contributor guide

jpvajda:
This PR updates the section on the SDK bundle to offer more context of how we do updates to documentation as it relates to the SDK bundle from One Core.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,7 +46,6 @@
     - [Step 1: Update the release number in `gatsby-config`](#step-1-update-the-release-number-in-gatsby-config)
     - [Step 2: Add any new APIs or components to our constants list](#step-2-add-any-new-apis-or-components-to-our-constants-list)
     - [Step 3: Add any new APIs or components to the navigation](#step-3-add-any-new-apis-or-components-to-the-navigation)
-    - [Step 4: Misc Info](#step-4-misc-info)
 
 <!-- /TOC -->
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -390,38 +391,40 @@ A good metric:
...
 
 ### Step 3: Add any new APIs or components to the navigation
 
 If there are new APIs or components, we will want to list them in the navigation
 so that a user can easily discover them. [Add an entry to `nav.yml`](https://github.com/newrelic/developer-website/blob/develop/src/data/nav.yml)
 to get the new API/component in the nav.
+
+### Step 4: Misc Info
+
+- This [file](https://github.com/newrelic/developer-website/blob/develop/plugins/gatsby-source-newrelic-sdk/src/navigationApi.js) contains the manually generated navigation API this is currently missing from the API.
```
jerelmiller, line 428: I'd hesitate to include all of this info here, mostly because the individual files may have a lot of churn and a lot of this stuff is internal implementation details. Maintaining the docs each time we make a change could get tiresome.

For example, this particular line about the navigation API is only useful while we need to polyfill that value. As soon as the platform provides documentation for it, we will then be able to go delete this.

I'll leave it up to you to decide whether its worth it, but just know this might be a bit volatile and could get out of date super quickly.

CONTRIBUTING.md
```diff
@@ -390,38 +391,40 @@ A good metric:
...
+- [Here](https://github.com/newrelic/developer-website/blob/develop/plugins/gatsby-source-newrelic-sdk/src/getPropTypes.js#L123-L128) is how prop type functions are extracted.
```
jerelmiller, line 429: If we decide to keep this, you might consider getting a permalink to these lines of code instead of pointing to whats in `develop`. If we make any changes to this file that change the lines where this functionality appears, or if we add additional functionality, which would expand the number of lines that are used (which I would expect if we are going to fix some of the bugs mentioned in #1374), then this link is going to point to something that makes no sense.

Again, if it feels valuable enough, then keep it, but this may get out of date pretty quickly as the functionality is updated.

CONTRIBUTING.md
```diff
@@ -390,38 +391,40 @@ A good metric:
...
-as that will contain only 3rd party SDK documentation.
+> Once we have a 3rd party bundle automatically built for us, we should no longer need this step as that will contain only 3rd party SDK documentation. That is an open request to the NR One Core Team.
 
 ### Step 3: Add any new APIs or components to the navigation
 
 If there are new APIs or components, we will want to list them in the navigation
 so that a user can easily discover them. [Add an entry to `nav.yml`](https://github.com/newrelic/developer-website/blob/develop/src/data/nav.yml)
 to get the new API/component in the nav.
+
+### Step 4: Misc Info
```
jerelmiller, line 426: Final though on this section, if nothing else, I'd just point this to the top-level local plugin folder and mention that the SDK documentation is generated from this plugin. Any additional "misc info" will be spread throughout other parts of the plugin, which may or may not change over time.

jerelmiller (commented):
I'll leave it up to you to decide whether the new section is useful or not, but just wanted to point out that the code could have a lot of churn and could cause this to get out of date pretty quick. I'd rather stick to high-level stuff (like pointing to the root folder) which is less likely to change if possible.

Again, I'll leave it up to you to decide what is best here!

timglaser (approved):

Issue #1349, opened 2021-05-27
[UI: Pages] Add mobile layout for search/landing page


roadlittledawn:
## Summary

We want to add a mobile / responsive layout for the pack search / landing page.

We don't have design specs yet for this, but we can make some reasonable assumptions. We'll likely want to stack all the elements, with search input / filters at the top and a list / condensed view of the packs.

Since left nav is TBD for this page, we may want to stack at either 1240px (when main and right sidebar stack) or 760px (when nav is hidden and accessible via hamburger menu in header).

## Acceptance criteria
- [x] On our mobile breakpoint, page elements stack with search input and filters at the top
- [x] The list of packs is more condensed so a user can scroll and quickly see all results

## Helpful links
- https://github.com/newrelic/developer-website/issues/1328
- [docs data dictionary page](https://github.com/newrelic/docs-website/blob/develop/src/pages/attribute-dictionary.js)
- [Theme Layout component](https://github.com/newrelic/gatsby-theme-newrelic/blob/develop/packages/gatsby-theme-newrelic/src/components/Layout/Layout.js) has some responsive-ness built-in

zstix:
I'm closing this since it looks like the related PR got merged in. Lookin' good!

PR #1378, opened 2021-06-07
Make o11y pack landing more responsive

roadlittledawn:
## Description

Adds more responsive-ness to landing page, especially form controls / filters. Adds some custom form components for more control over styling and behavior. Very similar to those used in docs site's data dictionary page.

The grid view is already responsive.

## Reviewer notes
* we may want to further refactor how filters work with state and URL params. i noticed that if you load a page with query params, the defaults are applied, then state is updated from query params.

Closes https://github.com/newrelic/developer-website/issues/1349

we may need to wait or merge the work for [list view](https://github.com/newrelic/developer-website/pull/1371) && [adding logos for featured images](https://github.com/newrelic/developer-website/pull/1372)

### DIFF
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -1,0 +1,94 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css } from '@emotion/react';
+
+const Select = ({ disabled, ...props }) => (
+  <div
+    css={css`
+      height: 28px;
+      display: grid;
+      grid-template-areas: 'select';
+      align-items: center;
+      width: 100%;
+      min-width: 15ch;
+      border: 1px solid var(--color-neutrals-500);
+      border-radius: 0.25rem;
+      padding: 0.25rem 0.5rem;
+      cursor: pointer;
+      position: relative;
+
+      ${disabled &&
+      css`
+        cursor: not-allowed;
+        opacity: 0.5;
+
+        .light-mode & {
+          background-color: var(--color-neutrals-100);
+        }
+      `}
+
+      .dark-mode & {
+        border: 1px solid var(--color-dark-300);
+        background-color: var(--color-dark-300);
+      }
+
+      &::after {
+        content: '';
+        grid-area: select;
+        justify-self: end;
+        width: 0.5rem;
+        height: 0.25rem;
+        clip-path: polygon(100% 0%, 0 0%, 50% 100%);
+        background-color: var(--color-neutrals-700);
+
+        .dark-mode & {
+          background-color: var(--color-dark-700);
+        }
+      }
+    `}
+  >
+    <select
+      disabled={disabled}
+      css={css`
+        grid-area: select;
+        appearance: none;
+        background-color: transparent;
+        border: none;
+        padding: 0 1em 0 0;
+        margin: 0;
+        width: 100%;
+        font-family: inherit;
+        font-size: 14px;
+        cursor: inherit;
+        line-height: inherit;
+        outline: none;
+
+        &:focus + .focus-ring {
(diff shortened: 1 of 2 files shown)


src/components/Select.js
```diff
@@ -0,0 +1,94 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css } from '@emotion/react';
+
+const Select = ({ disabled, ...props }) => (
+  <div
+    css={css`
+      height: 28px;
```
zstix, line 8: We might want to consider using `rem` for this so that it will scale with font-size settings.

```suggestion
      height: 1.75rem;
```

src/components/Select.js
```diff
@@ -0,0 +1,94 @@
...
+      display: grid;
+      grid-template-areas: 'select';
+      align-items: center;
+      width: 100%;
+      min-width: 15ch;
```
zstix, line 13: Do we need to set this? Even though `ch` is [mostly supported](https://caniuse.com/ch-unit), it's not very widely used (read: used at all) in our codebase.

src/components/Select.js
```diff
@@ -0,0 +1,94 @@
...
+        background-color: var(--color-dark-300);
+      }
+
+      &::after {
+        content: '';
+        grid-area: select;
+        justify-self: end;
+        width: 0.5rem;
+        height: 0.25rem;
+        clip-path: polygon(100% 0%, 0 0%, 50% 100%);
```
zstix, line 41: Out of curiosity, what are you trying to achieve here? [css-shapes](https://caniuse.com/css-shapes) is semi-supported.

src/components/Select.js
```diff
@@ -0,0 +1,94 @@
...
+        background-color: transparent;
+        border: none;
+        padding: 0 1em 0 0;
+        margin: 0;
+        width: 100%;
+        font-family: inherit;
+        font-size: 14px;
+        cursor: inherit;
+        line-height: inherit;
+        outline: none;
```
zstix, line 64: As much as I like this style, it's an [accessibility issue](http://www.outlinenone.com/) to remove `outline`:
```suggestion
```

zstix (commented):
I had a couple minor comments.

src/components/Select.js
roadlittledawn: oh we use this for the data dictionary filter select dropdowns as well. we create a custom select dropdown `div` that looks like the default, but this way we have a bit more control over styling / arrow.

src/components/Select.js
roadlittledawn: from what i understand it's useful in setting widths of input fields where you know the minimum number of characters you need to support.

we do use it in the [`Select`](https://github.com/newrelic/docs-website/blob/develop/src/components/Select.js) component which is used on the attribute dictionary page ¯\_(ツ)_/¯

roadlittledawn: i removed the `min-width` for now, but the `grid-template-areas` and `align-items` are necessary for the grid based div that emulates the default `select` styling,[ like i mentioned here](https://github.com/newrelic/developer-website/pull/1378#discussion_r647525575)

src/components/Select.js
roadlittledawn: that said, we could probably achieve most of what we want without the extra `div`, but the caret is a little thicker (in Chrome anyway) and may vary across browsers.

@zstix i'm kinda torn. should we ditch some of the extra-ness here and go for simplicity OR follow that pattern that gives us a little extra control (seemingly mostly for the caret)

zstix (approved):

PR #1383, opened 2021-06-08
add standard intro text to tabs pages

nr-kkenney:
### Screenshots

Screen Shot 2021-06-09 at 11 24 59 AM
Screen Shot 2021-06-09 at 11 25 06 AM
Screen Shot 2021-06-09 at 11 25 11 AM

### DIFF
--- a/src/templates/ObservabilityPackDetails.js
+++ b/src/templates/ObservabilityPackDetails.js
@@ -7,6 +7,7 @@ import PageLayout from '../components/PageLayout';
 import Tabs from '../components/Tabs';
 import { Layout, PageTools, Button } from '@newrelic/gatsby-theme-newrelic';
 import ImageGallery from '../components/ImageGallery';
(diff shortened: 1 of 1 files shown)


src/templates/ObservabilityPackDetails.js
```diff
@@ -78,7 +79,37 @@ const ObservabilityPackDetails = ({ data, location }) => {
                 <h3>Description</h3>
                 <p>{pack.description}</p>
               </Tabs.Page>
+              <Tabs.Page id="dependencies">
+                <Intro
+                  css={css`
+                    margin-bottom: 16px;
```
LizBaker, line 85: should we just add margin bottom to the intro component? seems more likely it would be wanted, and if it was ever going to the the bottom component then someone could specify no margin 🤷‍♀️

LizBaker (approved):

PR #1384, opened 2021-06-09
fix: Search onClear functionality

rudouglas:
`X` button now appears when there's a value in input, and clicking it clears the value

The `SearchInput` component requires `value` and `onClear` to be present to display the `X` button


LizBaker (approved):
🐑 it!

Issue #1389, opened 2021-06-08
fix the broken logo image for the couchbase logo


jpvajda:
### Issue 

We saw a few cases where the provided svg. for the logo defined for a pack isn't rendering correctly in this case we should default to the new relic default svg logo. we looked into a way perhaps detect if an SVG won't render and default to new relic logo, but we decided it's better to focus on putting some validations in place to ensure the logo file is in the correct svg format.

### Steps to reproduce 

- Go to https://observabilitypacks.gatsbyjs.io/observability-packs/?filter=Anything&sort=Alphabetical 
- look at the couchbase logo 🛋️ 

<img width="372" alt="Screen Shot 2021-06-08 at 5 32 24 PM" src="https://user-images.githubusercontent.com/1395158/121270781-94560800-c87f-11eb-8a63-7da647b75daa.png">

### Acceptance Criteria

- [x] ~~If an SVG doesn't render, show the default image for the new relic logo~~. - won't do
- [ ] fix the broken image for couchbase.

nr-kkenney:
@jpvajda for this particular image, it isn't loading correctly because the url in the data we have isn't correct. 
it currently is `https://raw.githubusercontent.com/newrelic/newrelic-observability-packs/v0.3.1/packs/couchbase/logo.png` but should be `https://raw.githubusercontent.com/newrelic/newrelic-observability-packs/v0.3.1/packs/couchbase/logo.svg`.

i had tried implementing something that would check if in image exists, but after running into some issues & talking with @zstix, it sounds like we should not implement this since the other images in the dev site don't do this, and this particular validation would be better implemented when the pack is first added (at the PR or API level).

is it cool if we close this out for now and make a note of it in the repo automation ticket?

jpvajda:
Yes. that works we are discussing the validations  mentioned in this issue with the virtuso team tomorrow https://github.com/newrelic/newrelic-observability-packs/issues/53 

I can add this into the specifications

jpvajda:
This issue got me thinking should we only accept .svgs? I know they are harder to produce but they are way more flexible to work with a UI. We currently are thinking of allowing .pngs and .jpegs. 🤔

jpvajda:
How do we fix the issue with referencing the wrong image url? we seem to be point to .png in the UI?

jpvajda:
re-opening until we figure out how to fix the broken image for couchbase as it's point to the wrong file.

nr-kkenney:
[...]

PR #1391, opened 2021-06-09
use default image if pack logo not available

nr-kkenney:
Resolves: #1389


nr-kkenney:
closing this PR per my comment in #1389

PR #1393, opened 2021-06-09
Observability Pack fetch workflow

### LINKED ISSUE #1356: Github action to fetch and check-in observability packs
## Description
Observability pack endpoints are now stitched into NerdGraph!

Since there are security concerns with making the observability pack API (the thing stitched into NerdGraph) public, and we don't want to check an API key into this repository, we have decided that the best path forward (at the moment) is to create an action that uses an API key stored in the workflow environment to query NerdGraph and pull down all the observability packs. To reduce impact on our other workflows, we should run this on an interval (rather than on all builds).

## Acceptance Criteria
- [x] Add API key to Github actions environment
- [x] Query NerdGraph to get _all_ observability packs
- [x] Save all packs as a JSON file
- [x] Check the JSON file into `main` (which should trigger a deploy)
- [x] Run the action on regular intervals
- [x] Write documentation for workflow
- [x] Write tests for the workflow

aswanson-nr:
## Description

Adds a workflow and associated script to pull down and commit Observability Packs to the `main` branch.

I did not add any unit tests, since there is minimal logic to actually test. If anyone sees anything that's worth breaking out and testing please let me know.

Closes #1356


.github/workflows/fetch-observability-packs.yml
```diff
@@ -0,0 +1,100 @@
...
+          key: ${{ runner.os }}-node-modules-${{ hashFiles('**/yarn.lock') }}
+
+      - name: Install dependencies
+        if: steps.yarn-cache.outputs.cache-hit != 'true'
+        run: yarn install --frozen-lockfile
+
+      - name: Fetch observability packs
+        run: yarn run fetch-observability-packs
+        env:
+          NR_GQL_URL: 'https://api.newrelic.com/graphql'
```
zstix, line 41: Super nitpicky, but would you mind renaming this to `NR_API_URL` to stay consistent with the other variable we're setting on the environment?

.github/workflows/fetch-observability-packs.yml
```diff
@@ -0,0 +1,100 @@
...
+            console.log("Result:", result)
+
+      - name: Commit changes
+        id: commit-changes
+        run: |
+          git config --local user.email "${{ env.BOT_EMAIL }}"
+          git config --local user.name "${{ env.BOT_NAME }}"
+          git add ./src/data/observability-packs.json
+          git diff-index --quiet HEAD ./src/data/observability-packs.json || git commit -m 'chore(observability-packs): updated observability packs'
+          echo "::set-output name=commit::true"
```
zstix, line 69: If this step fails (perhaps because there isn't any new information to commit), will that cause the remaining steps to error out? Just want to double-check that we're not opening the door to disabling the branch protection and failing before we can reenable it.

scripts/actions/fetch-observability-packs.js
```diff
@@ -0,0 +1,115 @@
...
+const fetchPacks = async (queryString, url, token) => {
+  try {
+    const results = await fetch(url, {
+      method: 'post',
+      body: JSON.stringify({ query: queryString }),
+      headers: {
+        'Content-Type': 'application/json',
+        'Api-Key': token,
+      },
+    }).then((res) => res.json());
```
zstix, line 66: We might want to consider avoiding `await` style syntax with `then` style syntax for promises.

An alternative would be to do something like:
```js
const res = await fetch();
const results = await res.json();
```

That would also allow us to more easily debug the response (e.g. log out the response code).

scripts/actions/fetch-observability-packs.js
```diff
@@ -0,0 +1,115 @@
...
+
+    if (results.errors) {
+      console.log('GRAPHQL Errors:', JSON.stringify(results.errors, null, 2));
+    }
```
zstix, line 70: If there are `results.errors` we probably want to drop out of this function (either by `return` or `throw`ing).

scripts/actions/fetch-observability-packs.js
```diff
@@ -0,0 +1,115 @@
...
+
+    return results.data.docs.openInstallation.observabilityPackSearch.results;
```
zstix, line 72: Just to be defensive, we might want to use something like [`_.get`](https://lodash.com/docs/4.17.15#get) (if we have lodash available) to ensure we don't hit an error here. Maybe the `catch` is sufficient tho, I dunno.

zstix (commented):
I had a few comments, but they're all pretty minor. Nice work!!

.github/workflows/fetch-observability-packs.yml
aswanson-nr: The jobs that reenable branch protection will always run, no matter what the outcome of this job is.

aswanson-nr: It has the `if: always()` setting

scripts/actions/fetch-observability-packs.js
aswanson-nr: My understanding of GraphQL is that it can still return partial information even if part of the query errors out. The API currently returns errors for the description field being blank on dashboards, but still returns the rest of the fields. Do you think it should just error out in that case?

scripts/actions/fetch-observability-packs.js
aswanson-nr: I was planning on letting the `try/catch` handle that case, it will catch a `cannot access property of undefined` error. I also don't mind being more explicit here

.github/workflows/fetch-observability-packs.yml
aswanson-nr: Done!

scripts/actions/fetch-observability-packs.js
aswanson-nr: Done!

scripts/actions/fetch-observability-packs.js
aswanson-nr: @zstix Do you think it's worth pulling in lodash?

scripts/actions/fetch-observability-packs.js
zstix: If the API changes and this query is no longer accurate, then we run the risk of hitting a hard-to-track-down error. For example, this query will return errors and _not_ contain `data.docs`:

```
{
  docs {
    openInstallation {
      observabilityPackSearch {
        results {
          observabilityPacks {
            taco
          }
        }
      }
    }
  }
}
```

I suppose it's not _critical_ that we throw an error here. If we don't then we _should_ consider checking the properties exist before returning (the line below this).

scripts/actions/fetch-observability-packs.js
zstix: We probably don't need _all_ of lodash. If we just pull in `get` itself, then we're only looking at adding [2 - 4kb](https://bundlephobia.com/package/lodash.get@4.4.2) to our bundle.

zstix: https://www.npmjs.com/package/lodash.get

aswanson-nr: Done!

scripts/actions/__tests__/fetch-observability-packs.test.js
```diff
@@ -0,0 +1,123 @@
...
+              }
+            }
+          }
+        }
+      }
+    };
+    fetch.mockResolvedValueOnce({
+      ok: true,
+      json: jest.fn(() => Promise.resolve(apiReturnValue))
+    });
```
zstix, line 48: 💯  yay tests!

I _think_ this could be simplified to:

```suggestion
    mockAPIResponse(apiReturnValue);
```

zstix: Same below.

scripts/actions/__tests__/fetch-observability-packs.test.js
```diff
@@ -0,0 +1,123 @@
...
+          }
+        }
+      }
+    };
+    fetch.mockResolvedValueOnce({
+      json: jest.fn(() => Promise.resolve(apiReturnValue))
+    });
+
+    await fetchObservabilityPacks(fakeGqlQuery, fakeAPIURL, fakeToken);
+    expect(fs.writeFileSync.mock.calls.length).toBe(0);
```
zstix, line 106: Just thinking out loud, but I think this could be more clear with something like

```js
expect(fs.writeFileSync).not.toHaveBeenCalled();
```

zstix: Same below.

scripts/actions/fetch-observability-packs.js
```diff
@@ -0,0 +1,137 @@
...
+ * @param {String} url the New Relic API endpoint
+ * @param {String} token a New Relic API token
+ **/
+const main = async (query, url, token) => {
+  const results = await fetchPacks(query, url, token);
+
+  if (results) {
+    const packs = results.observabilityPacks;
+    console.log(`Found ${packs.length} packs.`);
+    writePacks(PACKS_FILE_PATH, packs);
```
zstix, line 119: Out of curiosity (and this is more of a pondering question than a criticism), is there any value in moving this into a separate `writePacks` function instead of doing the following:
```suggestion
    fs.writeFileSync(PACKS_FILE_PATH, JSON.stringify(packs, null, 2));
```

zstix (commented):
A couple of suuuper minor comments / questions. I'm approving this tho cause it looks 🔥

zstix (approved):

scripts/actions/fetch-observability-packs.js
aswanson-nr: Great point, I think I originally planned to have more output, so I moved it to a function. Since all it does is log and write the file, I removed the function.

scripts/actions/__tests__/fetch-observability-packs.test.js
aswanson-nr: Done!

scripts/actions/__tests__/fetch-observability-packs.test.js
aswanson-nr: I actually meant to get rid of this function since I need very different response objects for each test! The latest commit removes it.

aswanson-nr:
@zstix Sorry to request your review again! I addressed your comments and resolved any merge conflicts

zstix (approved):

Issue #1385, opened 2021-06-08
Fade the gallery image on the right hand side


jpvajda:
###  Issue

The design called for a fading gradient effect on the pack gallery to soften the edge of the last image. We'll want to do that.

### mock up
<img width="938" alt="Screen Shot 2021-06-08 at 6 12 06 PM" src="https://user-images.githubusercontent.com/1395158/121273365-578d0f80-c885-11eb-893f-ac2d006eac67.png">

### Acceptance Criteria

- [ ] Implement the fading effect on the last image in the gallery

PR #1395, opened 2021-06-10
Add fade to image gallery

LizBaker:
## Description

This adds a fade to the right side of the image gallery to suggest there is more to scroll. I'm not sure if we want to add animation and get rid of the fade when you reach the last image 🤷‍♀️ but here's a first pass 🎉 

Closes https://github.com/newrelic/developer-website/issues/1385

## Screenshot(s)
Screen Shot 2021-06-09 at 5 51 29 PM
Screen Shot 2021-06-09 at 5 51 41 PM

### DIFF
--- a/src/components/ImageGallery/ImageGallery.js
+++ b/src/components/ImageGallery/ImageGallery.js
@@ -52,6 +52,7 @@ const ImageGallery = ({ images, className }) => {
           white-space: nowrap;
           overflow-x: auto;
           &:after {
+            pointer-events: none;
             content: '';
             position: absolute;
             background: linear-gradient(
--- a/src/components/ImageGallery/ImageGallery.js
+++ b/src/components/ImageGallery/ImageGallery.js
@@ -57,7 +57,7 @@ const ImageGallery = ({ images, className }) => {
             position: absolute;
             background: linear-gradient(
               to right,
-              transparent 75%,
+              transparent 87%,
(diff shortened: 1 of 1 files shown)


aswanson-nr (approved):

jpvajda:
I provided the following ux feedback to @LizBaker in slack, also adding here.

1. in Firefox, once the scroll bar disappears you can't scroll the gallery, though in chrome this isn't an issue

2. in light mode the fade is pretty intense

3.  I think each image should be clickable and open the image in a new tab so people can view it at a larger scale., --> might be a bug in css

4 I think the fade on the last image should not be present

jpvajda (changes requested):
see my last comment for suggestions...

LizBaker:
@jpvajda fixed the click-ability issue and reduced the fade. I couldn't reproduce the issue in firefox but that may be something that was fixed by fixing the click through. I also didn't come up with a good solution for getting rid of the fade at the end of the scroll so maybe we can make a separate ticket for that?

jpvajda:
@LizBaker Thanks for working on this I still feel the light-mode fade is a little `rough`, but I know CSS can take hours to get something like this right. 🤣  

Why don't we merge this in and I'll file a Nit to make the fade better on light mode, and also try to remove the fade on the last image.

aswanson-nr (approved):

PR #1399, opened 2021-06-11
feat: add install button and instrument

rudouglas:
Adds the Install Button component to the Pack Details page and instruments it with NR Browser and Tessen

### DIFF
--- a/src/templates/ObservabilityPackDetails.js
+++ b/src/templates/ObservabilityPackDetails.js
@@ -19,16 +19,16 @@ const ObservabilityPackDetails = ({ data, location }) => {
   const pack = data.observabilityPacks;
   const tessen = useTessen();
   const handleInstallClick = useInstrumentedHandler(
-    (pack) => {
-      tessen.track('observabilityPack', 'ObservabilityPackInstall', {
-        installPackName: pack.name,
-        installPackId: pack.id,
+    () => {
+      tessen.track('observabilityPack', 'packInstall', {
+        packName: pack.name,
+        packId: pack.id,
       });
     },
     {
-      actionName: 'oPackInstall',
-      installPackName: pack.name,
-      installPackId: pack.id,
+      actionName: 'packInstall',
+      packName: pack.name,
+      packId: pack.id,
     }
   );
   return (
@@ -46,7 +46,7 @@ const ObservabilityPackDetails = ({ data, location }) => {
             <InstallButton
               title={`Install ${pack.name}`}
               guid={pack.id}
-              onClick={() => handleInstallClick(pack)}
+              onClick={handleInstallClick}
             />
             <Tabs.Bar
               css={css`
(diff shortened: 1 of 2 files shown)


src/templates/ObservabilityPackDetails.js
```diff
@@ -24,9 +25,22 @@ const ObservabilityPackDetails = ({ data, location }) => {
               gap: 1rem;
             `}
           >
-            <Button variant={Button.VARIANT.PRIMARY} size={Button.SIZE.MEDIUM}>
-              Install
-            </Button>
+            <InstallButton
+              title="Install"
```
zstix, line 29: We might want to consider adding the pack name to the title:

```suggestion
              title={`Install ${pack.name}`}
```

rudouglas: I think that's a good call, but wasn't in the original design. Should we go for it? It's simples to change back should we need to?

src/templates/ObservabilityPackDetails.js
```diff
@@ -24,9 +25,22 @@ const ObservabilityPackDetails = ({ data, location }) => {
...
+              guid={pack.id}
+              onClick={() => {
+                tessen.track('observabilityPack', `ObservabilityPackInstall`, {
+                  installPackName: pack.name,
+                  installPackId: pack.id,
+                });
+                if (typeof window !== 'undefined' && window.newrelic) {
+                  window.newrelic.addPageAction('oPackInstall', {
```
zstix, line 37: I think we can take advantage of [`useInstrumentedHandler`](https://github.com/newrelic/gatsby-theme-newrelic/tree/develop/packages/gatsby-theme-newrelic#useinstrumentedhandler) here (to avoid doing extra checks for `window` and abstracting away the `addPageAction` stuff.

rudouglas: Ye I wasn't 100% sure how to use that within this context, maybe we can chat about this Monday too :D

zstix: I _think_ we can call `useInstrumentedHandler` to get the callback function used for the click event. Not 100% sure if this works, but here's the relevant bits of code I was thinking:

```jsx
// import the two hooks

import { useTessen, useInstrumentedHandler } from '@newrelic/gatsby-theme-newrelic';

// call the two hooks before the JSX code

const installPackData = { installPackName: pack.name, installPackId: pack.id };

const tessen = useTessen();

const handleInstallClick = useInstrumentedHandler(
  () => {
    tessen.track('observabilityPack', 'ObservabilityPackInstall', installPackData);
  },
  installPackData,
  'oPackInstall'
);

// use the handler from the hook in the JSX

<InstallButton
  onClick={handleInstallClick}
/>
```

rudouglas:
@zstix ready for review

zstix (approved):
🎸

PR #1400, opened 2021-06-11
Ruairi/instrumentation

rudouglas:
Instrumented components with NR Browser and Tessen


src/pages/observability-packs.js
```diff
@@ -108,7 +138,27 @@ const ObservabilityPacksPage = ({ data, location }) => {
...
+      tessen.track('observabilityPack', `ObservabilityPackClickThroughView`, {
+        oPackClickThroughView: view,
+      });
+      if (typeof window !== 'undefined' && window.newrelic) {
+        window.newrelic.addPageAction('oPackClickThroughView', {
+          oPackClickThroughView: view,
+        });
+      }
+    };
+    document.querySelectorAll('.pack-tile-instrument').forEach((item) => {
```
aswanson-nr, line 152: I could be wrong, but can we just add an `onClick` handler to the `Surface` component itself?

rudouglas: i did try that in a few different ways but it just seems to do nothing, and I couldn't figure out why 🤷 will campfire this monday to see if there are other options

rudouglas:
@zstix this has been updated :D

src/pages/observability-packs.js
```diff
@@ -47,7 +43,35 @@ const ObservabilityPacksPage = ({ data, location }) => {
   const [querySearch, setQuerySearch] = useQueryParam('search', StringParam);
   const [queryFilter, setQueryFilter] = useQueryParam('filter', StringParam);
   const [querySort, setQuerySort] = useQueryParam('sort', StringParam);
-
+  const [view, setView] = useState(VIEWS.GRID);
+  useInstrumentedData(
```
zstix, line 47: _super_ nitpcik: do you mind adding some blank lines between these hooks? It would be nice to match the style of the other components as closely as we can.

zstix (approved):
One minor comment, but this looks good to me!!

src/pages/observability-packs.js
rudouglas: yes of course, i've tidied them up and grouped similar ones, let me know if there's anything else needs massaging @zstix

zstix (approved):

PR #1409, opened 2021-06-16
Ruairi/upgrade sdk

rudouglas:


zstix:
Don't forget to link your issues 😄

src/components/FunctionDefinition.js
```diff
@@ -5,28 +5,36 @@ import { graphql } from 'gatsby';
...
-              <CodeDef.Comment text={param.description} />
-            </div>
-          ))}
-        </CodeDef.Block>
-      )}
-      {params.length > 0 && <CodeDef.Bracket>)</CodeDef.Bracket>}
-      <CodeDef.Operator> => </CodeDef.Operator>
-      <CodeDef.Type>{returnValue.type}</CodeDef.Type>
-    </CodeDef>
+    (params.length > 0 || returnValue.length > 0) && (
```
zstix, line 8: Totally optional, but you _could_ do a check for the inverse of these cases and do an early return, rather than wrapping whole component in this if statement:

```js
if (!params.length && !returnValue.length) {
  return;
}

return(
  // JSX goes here
);

zstix (changes requested):
Looks good (only on suggestion on style).

That said, it looks like you're checking in a `package-lock.json` file and we don't use `npm` for our dependency management.

Furthermore, there's changes checked into the `yarn.lock` file, but not changes to `package.json`. Might want to run `yarn` again to reset that.

zstix (approved):

Issue #1405, opened 2021-06-14
Public Catalog Nit Nit list `#2`


jpvajda:
This is a collection of nits to fix in the public catalog. This list 2 of our nits. :D 

### Nits

- [x] Fix the lightmode fade on the dashboard image file as it's a not a subtle fade.  See PR  #1395
- [x] The last dashboard image in a gallery should have no fade See  See PR  #1395
- [X] Change dependencies to requirements  See PR #1419
- [x] ~~The nav item for `Observability packs` should be highlighted when on a pack details page~~ - `won't do` see comment

### Need new tickets and design input from @danielgolden 

- [x] ~~add a back button on the olly pack details page so a user can get back to the grid view~~
- [x] change the display of dashboard images as they are really small in the gallery view @roadlittledawn has an idea for this and will make a #PR
- [x] Confirm if the Gallery component still works for how we display dashboard images, as if they are small they don't really offer much value
- [x] The fading image approach doesn't seem to work now that we implemented can we take that out?
- [x] Default (varies by browser) `select` styling for tab items for mobile (screenshot is iOS firefox) is a little weird (see [screenshot in comment](https://github.com/newrelic/developer-website/issues/1405#issuecomment-863581489)) https://github.com/newrelic/developer-website/issues/1421

roadlittledawn:
o11y pack detail page screenshot (iPhone X Firefox)
![IMG_8D7008DEA995-1](https://user-images.githubusercontent.com/2952843/122475482-93a52c00-cf79-11eb-86cf-97024189cc44.jpeg)

zstix:
> The nav item for Observability packs should be highlighted when on a pack details page

This would require a (somewhat significant) change to the way that the theme renders "active" navigation links. At the moment, the navigation item that's highlighted is the _current_ page. If we wanted, we could add the packs as their own navigation items in the nav, but I don't think that would scale well.

I think we should leave this _as is_. If we changed it as suggested, we would run the risk of the user thinking that they're _on_ the observability pack page and they might not realize that they can click on that link to get back.

jpvajda:
We are reviewing the UI feedback under the Needs New ticket section above with @danielgolden this week, once we have confirmation how to proceed I can make the necessary tickets for implementation. Moving to Sprint 7 as this ticket will carry over.

Issue #1388, opened 2021-06-08
add a Arrow button to move through the gallery so it's clear this is gallery component


jpvajda:
### Issue

The image gallery scrolls which is cool but it only shows it's scroll-able by showing the browser scroll bar. This works for people that have this enabled in a browser, but not for those that disable this browser feature.

### References

- Definitely look at the many gallery components that exist to get some ideas of this should work.. 

1. https://react-bootstrap.github.io/components/carousel/
2. https://reactjsexample.com/responsive-and-flexible-carousel-component-with-thumbnail-support/

If we need  design decision we can ask @danielgolden for a pattern we've used before.

### Acceptance Criteria

- [x]    add a Arrow button to move through the gallery so it's clear this is gallery component.
- [x]     On click, stop the gallery in the middle of an image
- [x]     As the images slide, keep the gradient fading on the right side of the image Related to: newrelic/developer-website#1385

zstix:
This has been completed in PR #1417

PR #1415, opened 2021-06-21
Add simple image slides with navigation

roadlittledawn:
## Description

Adds simple image slider (without the sliding animation for now) and arrows to navigate to next / prev image.

Does not show nav arrows if only one image

## Reviewer Notes

Wasn't sure what looked best for images either clipped on smaller screen or scaled down.

Related to:
- https://github.com/newrelic/developer-website/issues/1405
- https://github.com/newrelic/developer-website/issues/1388

## Screenshot(s)

https://user-images.githubusercontent.com/2952843/122829771-46d49480-d29c-11eb-98ce-c6d5450641e0.mp4

### DIFF
--- a/src/components/ImageSlider/ImageSlider.js
+++ b/src/components/ImageSlider/ImageSlider.js
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import React, { useState } from 'react';
 import { css } from '@emotion/react';
-import { Icon } from '@newrelic/gatsby-theme-newrelic';
+import { Button, Icon } from '@newrelic/gatsby-theme-newrelic';
 
 const ImageSlider = ({ images, height }) => {
   const [selectedImageIndex, setSelectedImageIndex] = useState(0);
@@ -29,8 +29,9 @@ const ImageSlider = ({ images, height }) => {
     >
       {images.length > 1 && (
         <>
-          <button
+          <Button
             onClick={handleClickPrev}
+            variant={Button.VARIANT.PLAIN}
             css={css`
               position: absolute;
               top: 38%;
@@ -42,9 +43,10 @@ const ImageSlider = ({ images, height }) => {
             `}
           >
             <Icon name="chevron-left" size="4rem" />
-          </button>
-          <button
+          </Button>
+          <Button
             onClick={handleClickNext}
+            variant={Button.VARIANT.PLAIN}
             css={css`
               position: absolute;
               top: 38%;
@@ -56,7 +58,7 @@ const ImageSlider = ({ images, height }) => {
             `}
           >
             <Icon name="chevron-right" size="4rem" />
-          </button>
+          </Button>
         </>
       )}
       <a
--- a/src/components/ImageSlider/ImageSlider.js
+++ b/src/components/ImageSlider/ImageSlider.js
@@ -62,7 +62,11 @@ const ImageSlider = ({ images, height }) => {
         </>
       )}
       <a
-        href={images && images.length > 0 ? images[selectedImageIndex] : ''}
+        href={
+          images && images.length > 0
+            ? images[selectedImageIndex]
+            : noImagePlaceholder
+        }
         target="_blank"
         rel="noreferrer"
         css={css`
--- a/src/components/ImageSlider/ImageSlider.js
+++ b/src/components/ImageSlider/ImageSlider.js
@@ -18,21 +18,48 @@ const ImageSlider = ({ images, height }) => {
         margin-bottom: 2rem;
       `}
     >
-      <button
-        onClick={handleClickNext}
+      {images.length > 1 && (
+        <button
+          onClick={handleClickNext}
+          css={css`
+            position: absolute;
+            top: 38%;
+            right: 0;
+            background: none;
+            color: var(--color-white);
+            border: none;
+            cursor: pointer;
+          `}
+        >
+          <Icon name="chevron-right" size="4rem" />
+        </button>
+      )}
+      <a
+        href={images[selectedImageIndex]}
+        target="_blank"
+        rel="noreferrer"
         css={css`
-          position: absolute;
-          top: 38%;
-          right: 0;
-          background: none;
-          color: var(--color-white);
-          border: none;
-          cursor: pointer;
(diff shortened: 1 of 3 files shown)


src/components/ImageSlider/ImageSlider.js
```diff
@@ -0,0 +1,110 @@
...
+  };
+
+  return (
+    <div
+      css={css`
+        position: relative;
+        margin-bottom: 2rem;
+      `}
+    >
+      {images.length > 1 && (
```
zstix, line 30: Rather than checking here, on the anchor tag, and on the image, it might make more sense to do a check if we have images and supply an early return:

```jsx
if (!images.length) {
  // empty state (no buttons, no link, etc.)
}

return (
  // ...other code
);
```

src/components/ImageSlider/ImageSlider.js
```diff
@@ -0,0 +1,110 @@
...
+            onClick={handleClickPrev}
+            variant={Button.VARIANT.PLAIN}
+            css={css`
+              position: absolute;
+              top: 38%;
+              left: 0;
+              background: none;
+              color: var(--color-white);
+              border: none;
+              cursor: pointer;
```
zstix, line 42: Totally optional, but you could probably cut down on a lot of the duplicate code by adding the `button` styles to the parent `div`. I think the only style that's unique to these buttons is the `left` / `right`.

src/components/ImageSlider/ImageSlider.js
```diff
@@ -0,0 +1,110 @@
...
+            `}
+          >
+            <Icon name="chevron-right" size="4rem" />
+          </Button>
+        </>
+      )}
+      <a
```
zstix, line 64: We might want to use [`ExternalLink`](https://github.com/newrelic/gatsby-theme-newrelic/tree/develop/packages/gatsby-theme-newrelic#externallink) for this.

src/components/ImageSlider/ImageSlider.js
```diff
@@ -0,0 +1,110 @@
...
+          margin-right: 1rem;
+        `}
+      >
+        <img
+          src={
+            images && images.length > 0
+              ? images[selectedImageIndex]
+              : noImagePlaceholder
+          }
+          alt="placeholder-text"
```
zstix, line 84: Perhaps this could be generated?

```jsx
alt={images && images.length ? `Image ${selectedImageIndex}` : 'No image available'}
```

src/components/ImageSlider/ImageSlider.js
```diff
@@ -0,0 +1,110 @@
...
+          css={css`
+            height: ${height}px;
```
zstix, line 86: Setting the height does some weird things on mobile:
Screen Shot 2021-06-21 at 2 54 04 PM

And it also looks like it squishes images a bit (depending on the aspect ratio). I wonder if there's a way we can avoid this 🤔 
Screen Shot 2021-06-21 at 2 54 59 PM

src/components/ImageSlider/ImageSlider.js
```diff
@@ -0,0 +1,110 @@
...
+          `}
+        />
+      </a>
+    </div>
+  );
+};
+
+ImageSlider.propTypes = {
+  images: PropTypes.array,
+  height: PropTypes.number,
```
zstix, line 104: _If_ we want to include a `height`, we should make it required.

src/components/ImageSlider/ImageSlider.js
```diff
@@ -0,0 +1,110 @@
...
+};
+
+const noImagePlaceholder =
+  'https://socialistmodernism.com/wp-content/uploads/2017/07/placeholder-image.png';
```
zstix, line 108: We might want to pull this into the repository (and optimize it).

src/components/ImageSlider/ImageSlider.js
```diff
@@ -0,0 +1,110 @@
...
+
+export default ImageSlider;
```
zstix, line 110: I noticed that this component is the only export in this directory. It seems like have been only making subdirectories under `components/` for components that have a few interrelated components (e.g. `Tabs` and `CodeDef`).

zstix (commented):
A couple of thoughts.

src/components/ImageSlider/ImageSlider.js
roadlittledawn: not sure i quite follow. i just wanted to only add nav arrows / dots if we have more than one image to navigate through. 

i wasn't 100% sure if i should account for no images here (and thus a placeholder image) or if it was safe to assume we would do that via the `src/templates/ObservabilityPackDetails.js` since we likely have to do some checks there to see what's in the data. ¯\_(ツ)_/¯

src/components/ImageSlider/ImageSlider.js
roadlittledawn: that parent `<div>` is for the entire slider container, so just added another `div` for these arrows and made them flex instead

src/components/ImageSlider/ImageSlider.js
roadlittledawn: i went for splitting and slicing the filename off the image URL

src/components/ImageSlider/ImageSlider.js
roadlittledawn: ah yep. i originally set it up that way thinking it would have multiple pieces and never looked back. thanks!

src/components/ImageSlider/ImageSlider.js
roadlittledawn: yeah i made the images `object-fit: contain` on mobile screens (smaller than 760px) because seeing the entire image (where you can click and see it in full resolution in another tab) seemed slightly better than seeing small portion of the image where you can't really discern what it is. 

i adjusted so it's full width and fixed height on mobile. looks okay, but TBH not sure what best option there is.

for the squishy-ness on those medium-large screens. i adjusted that in the new PR so it doesn't squish / distort so much.

roadlittledawn:
Closing in favor of animated version https://github.com/newrelic/developer-website/pull/1417

PR #1446, opened 2021-06-30
Use most recently stable version of SDK

LizBaker:
## Description

There's an issue with the `Transform data` example of the NrqlQueries component in our [sdk docs](https://developer.newrelic.com/components/nrql-queries/), this issue isn't present in the internal sdk docs site so I'm rolling us back to the version of the sdk they currently use

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -170,7 +170,7 @@ module.exports = {
     {
       resolve: 'gatsby-source-newrelic-sdk',
       options: {
-        release: 'release-2776',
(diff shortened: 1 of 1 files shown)


nr-kkenney (approved):

jpvajda:
As part of this PR we said we'd update [contributing guide](https://github.com/newrelic/developer-website/blob/develop/CONTRIBUTING.md#step-1-update-the-release-number-in-gatsby-config) for SDK to explain how to find the "right" sdk bundle

PR #1450, opened 2021-07-01
chore: add pack count to filter dropdown

rudouglas:
Screenshot 2021-07-01 at 16 57 22


jpvajda (approved):
LGTM We can change anything to all in a seperate Issue that I'll put in catch-all work for next sprint 👍

Issue #1447, opened 2021-06-30
Add a link to the pack's repo directory on the packs details page


jpvajda:
We'll want to highlight where a user can find the actually pack code on github so adding a clear CTA where the user can find the code would be nice.

### Acceptance Criteria

- [x] Work with `Kike Alonzo` on  design approach
- [x] leverage this pattern the  [OSS site pattern](https://opensource.newrelic.com/projects/newrelic/go-agent).
- [x] use the icon path to send the user to the correct pack directory.
- [x] IF no icon exists, send user to the [packs directory](https://github.com/newrelic/newrelic-observability-packs/tree/main/packs) OR send user to the correct pack directory.
- [x] Make the recommendation to add this field to pack to @jbeveland27 for a future release.

### Mock Up 

![#1447](https://user-images.githubusercontent.com/8354661/124607025-bc7c5b00-de6d-11eb-9d3c-9ec153a060f8.jpg)

kukukaka:
@jpvajda I think that's a great idea. I did a quick test trying to align that pattern to this use case. I don't have the GitHub icon on the library so I'm sharing it with the placeholder, but you can get the idea. Also, I'm not sure if we want to promote these pages to be editable so I didn't add that action, but let me know if we should add it.

![#1447](https://user-images.githubusercontent.com/8354661/124607025-bc7c5b00-de6d-11eb-9d3c-9ec153a060f8.jpg)

jpvajda:
We need to clarify where this button sends the users as the repo has packs in a directory... it's a slightly different flow then sending a user to  a dedicated  repository..

PR #1470, opened 2021-07-07
Add link to repo pack directory

nr-kkenney:
This PR adds a button to the pagetools section of a pack's page which, when clicked, takes the user to the packs main directory in the newrelic-observability-packs repo, or the packs/ directory for all packs (if the particular pack does not have a logo).

Resolves: #1447 

### Screenshots

Screen Shot 2021-07-07 at 5 02 25 PM


src/templates/ObservabilityPackDetails.js
```diff
@@ -32,6 +36,22 @@ const ObservabilityPackDetails = ({ data, location }) => {
       packId: pack.id,
     }
   );
+
+  const getPackUrl = () => {
```
zstix, line 40: We might want to put this in it's own util file. I could see it being difficult to track down where this logic is, if we need to update it in the future. Maybe something like `utils/get-pack-url.js` - what do ya think?

src/templates/ObservabilityPackDetails.js
```diff
@@ -32,6 +36,22 @@ const ObservabilityPackDetails = ({ data, location }) => {
...
+      'https://github.com/newrelic/newrelic-observability-packs/tree/main/packs';
+
+    if (!pack.logoUrl) {
+      return baseUrl;
+    }
+
+    /**
+     * logoUrl looks something like: 'https://raw.githubusercontent.com/newrelic/newrelic-observability-packs/v0.8.2/packs/couchbase/logo.svg'
+     */
+    const [packName, ..._] = pack.logoUrl.split('/').slice(-2);
```
zstix, line 51: You don't actually need to destructure the remaining arguments. If all you want is the first item you can just do this:

```suggestion
    const [packName] = pack.logoUrl.split('/').slice(-2);
```

zstix: Furthermore, if you're just trying to get the packname, you shouldn't need to destructure anything: just supply a second parameter to [`Array.slice`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice):
```suggestion
    const packName = pack.logoUrl.split('/').slice(-2, -1);
```

src/templates/ObservabilityPackDetails.js
```diff
@@ -32,6 +36,22 @@ const ObservabilityPackDetails = ({ data, location }) => {
...
+    return join(baseUrl, packName);
```
zstix, line 52: I think the original intention behind `path` was to join filepaths for file resources. We might want to reach for [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) instead. I think it more accurately describes what we're doing:

```js
const base = 'https://github.com/newrelic/newrelic-observability-packs/tree/main/packs';

const name = pack.logoUrl.split('/').slice(-2, -1);
const url = new URL(name, base);

return url.href;
```

zstix (changes requested):
My requests on the PR preview branch are all going to a `404`. Looks like the URL has some placeholder text in it: https://observabilitypacks-addlinktorepopackdirectory.gtsb.io/github.com/newrelic/newrelic-observability-packs/tree/main/packs/cassandra

Left a few other comments about code style.

Also, why was the lock file updated? 🤔

src/templates/ObservabilityPackDetails.js
aswanson-nr, line 51: Deconstruction allows you to just specify what you need, so if you just want the head, you only need to declare the `packName` variable.
```suggestion
    const [packName] = pack.logoUrl.split('/').slice(-2);
```

src/templates/ObservabilityPackDetails.js
aswanson-nr, line 52: The `path` module is exclusively for system file paths. The equivalent for URLs would be
```suggestion
    return new URL(packName, baseUrl).href;
```

src/templates/ObservabilityPackDetails.js
```diff
@@ -230,6 +250,28 @@ const ObservabilityPackDetails = ({ data, location }) => {
               }
             `}
           >
+            <PageTools.Section
+              css={css`
+                background-color: var(--divider-color);
+              `}
+            >
+              <div>
+                <Button
```
aswanson-nr, line 259: It could be worth adding some event tracking for this button click, just an idea!

aswanson-nr (dismissed):
Looks great overall! I just have a few requested changes and one idea that you definitely don't have to do.

src/templates/ObservabilityPackDetails.js
aswanson-nr: Resolving this since @zstix already covered it in his comments

src/templates/ObservabilityPackDetails.js
nr-kkenney: added

zstix (approved):

src/templates/ObservabilityPackDetails.js
nr-kkenney: moved into its own file 👍

src/templates/ObservabilityPackDetails.js
nr-kkenney: changed to use url, and shorter destructure

src/templates/ObservabilityPackDetails.js
nr-kkenney: swapped to url

nr-kkenney: swapped to url

Issue #1421, opened 2021-06-23
Default (varies by browser) select styling for tab items for mobile is displaying poorly


jpvajda:
the drop down selector style for the tab items varies by browser, we should ensure we are consistently presenting this text in a format that is readable. 

### Acceptance Criteria 

- [x] ensure we display the text in the drop down consistently on both light and dark modes in all browers for both mobile and desktop.
- [x] this may require some testing of chrome, firefox, and safari on desktop and mobile devices to check our consistency.

**I-phone Firefox** 

![IMG_8D7008DEA995-1](https://user-images.githubusercontent.com/2952843/122475817-0dd5b080-cf7a-11eb-9be1-cab7e5281a04.jpeg)

jpvajda:
we could use our dropdown component for this need, but we'll want to check the accessibility of that component and ensure that is accessible to all types of users.

zstix:
This seems to be an iOS issue: you can set the foreground color of the `<select>` element, but not the background color. If I recall correctly, all browsers on iOS use the same rendering engine as safari (webkit). We can just unset the colors on that browser / rendering engine.

PR #1471, opened 2021-07-08
Change tab select styling on ios browsers

roadlittledawn:
## Description

Checks if mobile device is iOS and if so adjusts to text color in the select to be more readable on the gray background

Closes https://github.com/newrelic/developer-website/issues/1421

## Screenshot(s)

IMG_CEB0044F5DB1-1


zstix (approved):
Nice!

Issue #1434, opened 2021-06-25
Reduce Tabs to 5 to align with platform UX guidelines on tabs


jpvajda:
* We need to have less Tabs, the platform UI is limited to 4 tabs we 7 total, let's try to get to 5 tabs. This needs design input before starting. 

### Ideas

  * Combine Visualizations and Nerd packs as a Viz is a nerdpack (_technically_)
  * Nerdpackss are out of scope for the MVP so we don't need to display this just yet
  * Put "Requirements" meta data  on the overview tab (_we need to confirm what will populate the requirements of a pack_)
  * Define what Requirements actually are, we assume it's based on the [instrumentation section of the config](https://github.com/newrelic/newrelic-observability-packs/blob/b80674e6594e5025b0803b644c585dd1df7f07bf/_template/config.yml#L22)

### Acceptance Criteria

- [X] Work with `Kike Alonzo` on a design and approach first 
- [x] Requirements tab is removed 
- [x] Instrumentation meta data appears in a new section below the `How to Use` right rail box.
- [x] Nerdpack tab is removed
- [x] Don't change synthetics to synthetic checks (that's from an old design)

### Mock Up
![#1434](https://user-images.githubusercontent.com/8354661/124609260-c30bd200-de6f-11eb-8d93-c15c504e6af9.jpg)

kukukaka:
@jpvajda I added Requirements and Support (it was there on some of the designs) on the overview sidebar, and replaces Nerpacks with Viz. It follows the pattern we have on some other places on the site, so I think it could be a good option. 

I used the instrumentation example you shared for the requirements, but I think we would need some help or further explanations to make it understandable by the user, right?

![#1434](https://user-images.githubusercontent.com/8354661/124609260-c30bd200-de6f-11eb-8d93-c15c504e6af9.jpg)

jpvajda:
I will confirm this approach before we implement.

jpvajda:
This is now ready @zstix so perhaps we can point it a standup this week

PR #1472, opened 2021-07-08
Reduce number of tabs in observability pack details

LizBaker:
## Description

This removes the requirements and nerdpacks tabs from the observability pack details page, and mocks the requirements info in the right rail. I also adjusted some mobile styles to avoid a scroll bar on the tabs as much as possible

Closes https://github.com/newrelic/developer-website/issues/1434

## Screenshot(s)
Screen Shot 2021-07-08 at 2 19 43 PM
Screen Shot 2021-07-08 at 2 19 31 PM

### DIFF
--- a/src/components/Tabs/Bar.js
+++ b/src/components/Tabs/Bar.js
@@ -58,6 +58,7 @@ const Bar = ({ children, className }) => {
       css={css`
         display: flex;
         width: 100%;
+        border-bottom: 1px solid var(--color-neutrals-300);
         margin-bottom: 1em;
         overflow: auto;
       `}
--- a/src/components/Tabs/Bar.js
+++ b/src/components/Tabs/Bar.js
@@ -58,7 +58,7 @@ const Bar = ({ children, className }) => {
       css={css`
         display: flex;
         width: 100%;
-        border-bottom: 1px solid var(--color-neutrals-300);
+        border-bottom: 1px solid var(--divider-color);
         margin-bottom: 1em;
         overflow: auto;
       `}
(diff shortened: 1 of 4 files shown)


src/components/Tabs/Bar.js
```diff
@@ -58,6 +58,7 @@ const Bar = ({ children, className }) => {
       css={css`
         display: flex;
         width: 100%;
+        border-bottom: 1px solid var(--color-neutrals-300);
```
aswanson-nr, line 61: This color is very bright in dark mode, maybe we could use something a little closer to the left rail separator.
Screen Shot 2021-07-08 at 3 59 28 PM

zstix, line 61: Can we also supply a color for the dark theme? I think we were using `var(--color-dark-300)` before.

zstix (changes requested):
Two 'lil things to update, but this looks great otherwise!

aswanson-nr (dismissed):
The border-bottom for the tabs looks a bit bright on darkmode, but it was like that before this change so feel free to ignore my comment

zstix:
I also think that this removed our mobile dropdown styles. Was that intentional?

Screen Shot 2021-07-08 at 13 02 42

src/components/Tabs/Bar.js
LizBaker: good catch, i'm trying out --divider-color which is also used by the right rail, do you think it's too subtle? also realized I'd lost the button underline when selected so I added that back

LizBaker:
@zstix we didn't have great mobile styles for the tabs imo, it was a scroll bar previously which I found awkward

zstix:
@LizBaker I think we had a browser-based dropdown (thus: https://github.com/newrelic/developer-website/issues/1421)

zstix (approved):
Ignore the comment about the dropdown, I forgot we were doing it based on user-agent and not device-width 🤦‍♂️

LizBaker:
@zstix ohhh gotcha, plus this just adds flex wrap past the 760 break point so i don't think it'll affect dropdown stuff when that's resolved 🤷‍♀️

Issue #1432, opened 2021-06-25
Enable empty tabs to be clickable to promote contributions


jpvajda:
This will require design and copy but the idea is, if a tab has no component, we'd allow the user to click on the tab and they could see a  CTA of  how they can contribute to the pack in the case if a pack lacks a dashboard, synthetic, alert, nerdpack. 

### Acceptance Criteria

- [x] Work with `Kike Alonzo` on our design approach to encourage contributions on these tabs
- [x] Work with @caylabudhdev on copy for this page - @jpvajda 
- [x] Content can be found in `Google Drive`>   `UIWRITE-1255: [Observability packs] Review & update copy for pack details `
- [x] Default empty state content is show on each tab 
- [x] view repo link would work the same as #1447

![#1432](https://user-images.githubusercontent.com/8354661/124613211-4844b600-de73-11eb-9622-d238586cb6f6.jpg)

kukukaka:
Hi!
I need some help with the copy, but here there is an early proposal for the empty case. I'd prefer to have a nice illustration there, but I added an icon as it would require more time to do it. Let me know what you think @jpvajda.

![#1432](https://user-images.githubusercontent.com/8354661/124613211-4844b600-de73-11eb-9622-d238586cb6f6.jpg)

jpvajda:
view repo link would the same as #1447

jpvajda:
I'm confirming with Cayla what content we can use for empty state. I believe there is a draft of it in the this doc: Content can
`Google Drive`> `UIWRITE-1212: [Observability packs] Write tab intros for the pack details pages`

PR #1477, opened 2021-07-09
Observability Packs default tab state

aswanson-nr:
## Description

This PR removes the `disabled` attribute from tabs that do not have any content and adds a default state for empty tabs. I also broke the rendering of these tabs out into functions because the JSX was getting gnarly.

I reduced the border width of the empty state content to better align with the right rail.

* Closes [Enable empty tabs to be clickable to promote contributions](https://github.com/newrelic/developer-website/issues/1432)

## Screenshot(s)
Screen Shot 2021-07-09 at 4 06 12 PM
Screen Shot 2021-07-09 at 4 06 21 PM
Screen Shot 2021-07-09 at 4 08 50 PM


aswanson-nr:
Just need final approval of the text content from @jpvajda and this should be good to go.

src/templates/ObservabilityPackDetails.js
```diff
@@ -124,124 +274,25 @@ const ObservabilityPackDetails = ({ data, location }) => {
...
-                  Ut enim ad minim veniam, quis nostrud exercitation ullamco
-                  laboris nisi ut aliquip ex ea commodo consequat. Duis aute
-                  irure dolor in reprehenderit in voluptate velit esse cillum
-                  dolore eu fugiat nulla pariatur. Excepteur sint occaecat
-                  cupidatat non proident, sunt in culpa qui officia deserunt
-                  mollit anim id est laborum.
-                </Intro>
+                {pack.synthetics
+                  ? renderSynthetics()
+                  : emptyStateContent(pack, 'Synthetics checks')}
```
rudouglas, line 290: Just wondering if this should be `Synthetics checks` or just `Synthetics`, I know we recently changed that for the tab itself

src/templates/ObservabilityPackDetails.js
```diff
@@ -37,6 +37,169 @@ const allowedElements = [
...
+        >
+          <Icon
+            name="fe-github"
+            css={css`
+              margin-right: 7px;
+            `}
+          />
+          View Repo
+        </Button>
+      </div>
```
rudouglas, line 198: This may be a nit or not, but the double `View Repo` button looks a little off
Screenshot 2021-07-12 at 15 10 33

rudouglas (approved):
LGTM apart from a couple potential nits

jpvajda:
This looks sooo good... 👌

src/templates/ObservabilityPackDetails.js
aswanson-nr: I thought this was a bit odd, but my understanding is this is the correct wording for the product.

src/templates/ObservabilityPackDetails.js
aswanson-nr: I agree that it's a bit odd to have the double button, especially on mobile. @jpvajda Do you have any thoughts on this? I think we can leave it for mobile nits

src/templates/ObservabilityPackDetails.js
```diff
@@ -37,6 +37,169 @@ const allowedElements = [
...
+        occaecat cupidatat non proident, sunt in culpa qui officia deserunt
+        mollit anim id est laborum.
+      </Intro>
+      {content}
+    </>
+  );
+};
+
+const renderAlerts = (alerts) => {
+  const alertContent = alerts.map((alert) => (
```
zstix, line 75: I think react might yell at us here for not providing a `key`.

src/templates/ObservabilityPackDetails.js
```diff
@@ -37,6 +37,169 @@ const allowedElements = [
...
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
+      tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim
+      veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea
+      commodo consequat. Duis aute irure dolor in reprehenderit in voluptate
+      velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat
+      cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id
+      est laborum.
+    </Intro>
+  );
+};
```
zstix, line 123: These all look fairly similar. Perhaps we could abstract this out into it's own component. Something like `ObservabilityPackEmptyState`?

zstix: Oh, I misunderstood, this is the _non-empty_ state? Either way, we might have little too much duplication in this file.

zstix (commented):
Couple of small comments. That said, I'm not 100% clear on what the purpose of this lorem ipsum stuff is for so they might not be applicable.

src/templates/ObservabilityPackDetails.js
aswanson-nr: I agree, I just didn't want to abstract it since the copy will be different for each state once that work gets done. Do you think it would be worth abstracting for now?

aswanson-nr: Specifically the changes in this PR https://github.com/newrelic/developer-website/pull/1478

src/templates/ObservabilityPackDetails.js
aswanson-nr: Updated!

nr-kkenney (approved):
site looks good to me. i think for the dashboards tab, i would prefer the images above the intro section, but i think thats its own thing

src/templates/ObservabilityPackDetails.js
aswanson-nr: @zstix Can you take a look at this now that the unique copy has gone in for each tab? I didn't have any ideas of the top of my head for abstracting a tab page since each one requires different text and portions of the pack itself.

zstix (approved):

Issue #1481, opened 2021-07-12
Surface how to `build your own Pack` in the public catalog


jpvajda:
It be cool to allow a  user to see a clear call to action regarding how they can build their own o11y packs, and guide them to our contributor guide.  We have a similar pattern we use within the product to encourage a user to build a nerdpack. 

### Crude Mock Up 👩‍🍳 

<img width="1379" alt="Screen Shot 2021-07-12 at 4 25 33 PM" src="https://user-images.githubusercontent.com/1395158/125364045-044c2800-e32f-11eb-957c-741f4860853f.png">

### Acceptance Criteria

- ~Obtain a final design from `kike alonzo`~
- [x] A Build your own pack tile is always the first tile visible 
- [x] On click, the user is taking to the Pack Repo.

zstix:
@jpvajda Are you cool if we mark this as complete (or push it to a later sprint)? We have completed the engineering work.

jpvajda:
@zstix sure, I filed this as a follow up 🎫 

https://github.com/newrelic/developer-website/issues/1485

PR #1484, opened 2021-07-14
Add build your own pack tile

LizBaker:
## Description

This PR is for adding a tile to the observability pack list that links to the pack repo to encourage folks to build their own

TODO
- [ ] confirm and implement design
- [ ] confirm and implement copy
- [x] adjust instrumentation for `build your own pack` clicks
- [x] decide whether to create a new component or stick with the manipulations to PackTile
      This issue is mainly stemming from the fact that the tile needs to externally link and is currently built for internal links

Closes https://github.com/newrelic/developer-website/issues/1481

### Current WIP
Screen Shot 2021-07-14 at 2 43 22 PM

### DIFF
--- a/src/components/PackImg.js
+++ b/src/components/PackImg.js
@@ -46,16 +46,12 @@ const PackImg = ({ className, logoUrl, packName }) => {
         background-repeat: no-repeat;
         background-position: center;
         background-origin: content-box;
+        display: flex;
+        justify-content: center;
+        align-items: center;
       `}
     >
-      <p
-        css={css`
-          text-align: center;
-          line-height: ${packAcronym.length < 4 ? '190%' : '300%'};
-        `}
-      >
-        {packAcronym}
-      </p>
+      <p>{packAcronym}</p>
     </div>
   );
 };
--- a/src/components/PackImg.js
+++ b/src/components/PackImg.js
@@ -1,16 +1,23 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/react';
-import DEFAULT_IMAGE from '../images/new-relic-logo.png';
+import DEFAULT_IMAGE from '../images/default-logo-background.svg';
 
 const PackImg = ({ className, logoUrl, packName }) => {
+  const [packAcronym, setPackAcronym] = useState('');
+
   const getPackNameAcronym = () => {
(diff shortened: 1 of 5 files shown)


src/pages/observability-packs.js
```diff
@@ -351,6 +352,24 @@ const ObservabilityPacksPage = ({ data, location }) => {
...
+          href="https://github.com/newrelic/newrelic-observability-packs"
+          target="_blank"
+          rel="noreferrer"
+          css={css`
+            text-decoration: none;
+          `}
+        >
+          <PackTile
+            css={css`
+              height: 100%;
```
zstix, line 365: I think we'll need to do something to address the position in the list view (right now it's doing some weird things).

src/pages/observability-packs.js
```diff
@@ -351,6 +352,24 @@ const ObservabilityPacksPage = ({ data, location }) => {
           `}
         `}
       >
+        <a
+          href="https://github.com/newrelic/newrelic-observability-packs"
+          target="_blank"
+          rel="noreferrer"
```
zstix, line 358: I _think_ we might be able to use the `ExternalLink` component for this. Not sure if it's worth the update, just wanted to mention it.

src/components/PackImg.js
```diff
@@ -34,56 +41,20 @@ const PackImg = ({ className, logoUrl, packName }) => {
...
-          position: relative;
-          top: 80%;
-          left: 83%;
-        `}
-      />
       <p
         css={css`
           text-align: center;
-          line-height: 150%;
+          line-height: ${packAcronym.length < 4 ? '190%' : '300%'};
```
zstix, line 54: Hopefully this covers the most common acronym lengths. We might be able to calculate a % based on the acronym length. Another possible option is to just wrap to a new line.

This is fine as-is, I'm just thinking out loud.

zstix (changes requested):
Lookin' good!

The only thing I'd like to see us update is the treatment of the new tile in list-view.

src/components/PackImg.js
aswanson-nr: This was removed in favor of using flexbox for centering

src/pages/observability-packs.js
aswanson-nr: This should be fixed now!

zstix (approved):

PR #1500, opened 2021-07-29
Switch install links to nr1 staging

roadlittledawn:
## Description

Switches install button links on pack details pages to point to staging-one.newrelic.com by default instead of using dev and using locally running nerdpack.

### DIFF
--- a/src/components/InstallButton.js
+++ b/src/components/InstallButton.js
@@ -6,7 +6,7 @@ import { NR1_LOGIN_URL } from '../data/constants';
 
 /** @param {string} packId */
 const createInstallLink = (packId) => {
-  const platformUrl = getPackNr1Url(packId, false); // FIXME: remove `true` when deployed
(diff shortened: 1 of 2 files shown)


src/components/InstallButton.js
```diff
@@ -6,7 +6,7 @@ import { NR1_LOGIN_URL } from '../data/constants';
 
 /** @param {string} packId */
 const createInstallLink = (packId) => {
-  const platformUrl = getPackNr1Url(packId, true); // FIXME: remove `true` when deployed
+  const platformUrl = getPackNr1Url(packId); // FIXME: remove `true` when deployed
```
LizBaker, line 9: shall we also remove the comment here?

roadlittledawn: good call. looked right past it. done!

LizBaker (approved):

PR #1507, opened 2021-08-04
Add temporary test cookie

LizBaker:
This (temporarily) adds a cookie on load of the futurehack page. We want to test out the use of a cookie to pass pack information to one.newrelic for installs as outlined here https://github.com/newrelic/newrelic-observability-packs/issues/127


zstix (approved):
🍪

Issue #1511, opened 2021-08-05
[Public Catalog] Create a “Data sources” tab for packs


jpvajda:
## Description
The [ObservabilityPackDetails](https://github.com/newrelic/developer-website/blob/feature/o11y-packs/src/templates/ObservabilityPackDetails.js) template in the Public Catalog will need to have a new tab labelled `Data sources`. This is were the new `documentation` info created as part of the [Directory Migration](https://github.com/newrelic/newrelic-observability-packs/issues/128) will be surfaced.

`NOTE:` This ticket is just for creating the new tab, populating that tab with `documentation` info is part of  #1516

<img width="1665" alt="integration_pack" src="https://user-images.githubusercontent.com/1395158/128577041-0e74776c-f8fa-405d-b56d-f7ca3939eb8d.png">

## Acceptance Criteria
- [ ] All packs should have a `Data sources` tab 
- [ ] This tab should follow the existing tab functionality we have in place for count display and visiblity of tabs with no content.

PR #1517, opened 2021-08-17
Add data sources tab

LizBaker:
Adds a data source tab and page with empty state
Closes https://github.com/newrelic/developer-website/issues/1511

## Screenshot(s)
Screen Shot 2021-08-17 at 3 07 13 PM

### DIFF
--- a/src/templates/ObservabilityPackDetails.js
+++ b/src/templates/ObservabilityPackDetails.js
@@ -226,6 +226,7 @@ const ObservabilityPackDetails = ({ data, location }) => {
             <Tabs.BarItem id="alerts" count={pack.alerts?.length ?? 0}>
               Alerts
             </Tabs.BarItem>
+            <Tabs.BarItem id="data-sources">Data Sources</Tabs.BarItem>
           </Tabs.Bar>
           <Layout.Content>
             <Tabs.Pages>
@@ -245,6 +246,9 @@ const ObservabilityPackDetails = ({ data, location }) => {
                 {pack.alerts
                   ? renderAlerts(pack)
                   : emptyStateContent(pack, 'alerts')}
+              </Tabs.Page>
+              <Tabs.Page id="data-sources">
+                {emptyStateContent(pack, 'data sources')}
               </Tabs.Page>
             </Tabs.Pages>
           </Layout.Content>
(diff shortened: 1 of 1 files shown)


src/templates/ObservabilityPackDetails.js
```diff
@@ -226,6 +226,7 @@ const ObservabilityPackDetails = ({ data, location }) => {
             <Tabs.BarItem id="alerts" count={pack.alerts?.length ?? 0}>
               Alerts
             </Tabs.BarItem>
+            <Tabs.BarItem id="data-sources">Data sources</Tabs.BarItem>
```
aswanson-nr, line 229: I think we want to include the count of data sources in this tab title, similar to the other tabs. Worth mentioning that this is the combination of the `instrumentation` and `documentation` fields

aswanson-nr (changes requested):

src/templates/ObservabilityPackDetails.js
LizBaker: I was wondering about this, it felt kind of weird to count those as individual items but i'm happy to add if that makes sense to you

rudouglas: Ye I agree just to keep it consistent with the design of the other tabs (though also think weird)

src/templates/ObservabilityPackDetails.js
```diff
@@ -226,7 +226,15 @@ const ObservabilityPackDetails = ({ data, location }) => {
             <Tabs.BarItem id="alerts" count={pack.alerts?.length ?? 0}>
               Alerts
             </Tabs.BarItem>
-            <Tabs.BarItem id="data-sources">Data sources</Tabs.BarItem>
+            <Tabs.BarItem
+              id="data-sources"
+              count={
+                (pack.instrumentation?.length ?? 0) +
+                (pack.documentation ? 1 : 0)
```
aswanson-nr, line 233: Sorry to toss another one on here! But the documentation field will also be an array

LizBaker: thank you 😂 I had it in my head that was a string

aswanson-nr (approved):

PR #1519, opened 2021-08-19
chore: add packUrl field

rudouglas:
This is to replace the current way of grabbing the URL to the pack in the repo from the logo url, and use the `packUrl` field in NerdGraph instead

Also removed the util function for parsing that url `get-pack-url.js`


scripts/actions/fetch-observability-packs.js
```diff
@@ -73,6 +74,7 @@ const fetchPacks = async (queryString, url, token) => {
     const results = await res.json();
 
     if (results.errors) {
+      console.log(results.errors);
```
zstix, line 77: It looks like we're `throw`ing, `catch`ing and calling `console.error` down below so we probably don't need this extra `console.log`.

```suggestion
```

src/templates/ObservabilityPackDetails.js
```diff
@@ -158,7 +157,7 @@ const emptyStateContent = (pack, tabName) => {
         <Button
           as={Link}
           variant={Button.VARIANT.PRIMARY}
-          to={getPackUrl(pack.logoUrl)}
+          to={pack.packUrl}
```
zstix, line 160: It looks like `packUrl` [is optional](https://source.datanerd.us/open-source/open-install-library-service/blob/master/src/graphql.schema.ts#L168). We might want to supply a fallback (e.g. the url for the quickstart repository) in the event that this url doesn't exist:

```js
// if it doesn't already exist as a constant...
const REPO_URL = "https://github.com/newrelic/newrelic-observability-packs";

const packUrl = pack.packUrl || REPO_URL;

// ...
to={packUrl}
```

zstix (commented):
A few comments we might want to address before bringing this work in.

src/templates/ObservabilityPackDetails.js
rudouglas: cool ye good to have a failsafe, would this also work:

```js
const { packUrl = REPO_URL } = pack;
```

rudouglas: i think it's probably a little less clear but just out of interest

aswanson-nr (dismissed):

src/data/constants.js
```diff
@@ -19,3 +19,6 @@ export const NR1_BASE_URL = 'https://staging-one.newrelic.com';
 
 export const NR1_PACK_DETAILS_NERDLET =
   'catalog-pack-details.catalog-pack-contents';
+
+export const REPO_URL =
+  'https://github.com/newrelic/newrelic-observability-packs';
```
zstix, line 24: Nitpick: `REPO_URL` is a little ambiguous (sorry, I should have provided a better suggestion). Since this is the URL for the Observability Packs / Quickstarts and not the current repository (developer-website), we should probably update this accordingly:
```suggestion
export const QUICKSTARTS_REPO =
  'https://github.com/newrelic/newrelic-observability-packs';
```

src/pages/observability-packs.js
```diff
@@ -19,6 +19,8 @@ import { useQueryParam, StringParam } from 'use-query-params';
 import { useDebounce } from 'react-use';
 import BUILD_YOUR_OWN from '../images/build-your-own.svg';
 
+const { REPO_URL } = require('../data/constants');
```
zstix, line 22:
```suggestion
const { QUICKSTARTS_REPO } = require('../data/constants');
```

src/pages/observability-packs.js
```diff
@@ -378,7 +380,7 @@ const ObservabilityPacksPage = ({ data, location }) => {
         `}
       >
         <ExternalLink
-          href="https://github.com/newrelic/newrelic-observability-packs"
+          href={REPO_URL}
```
zstix, line 383:
```suggestion
          href={QUICKSTARTS_REPO}
```

src/templates/ObservabilityPackDetails.js
```diff
@@ -18,10 +18,11 @@ import ImageGallery from '../components/ImageGallery';
 import Intro from '../components/Intro';
 import InstallButton from '../components/InstallButton';
 import ImageSlider from '../components/ImageSlider';
-import getPackUrl from '../utils/get-pack-url';
 import Markdown from '../components/Markdown';
 import pluralize from 'pluralize';
 
+const { REPO_URL } = require('../data/constants');
```
zstix, line 24:
```suggestion
const { QUICKSTARTS_REPO } = require('../data/constants');
```

src/templates/ObservabilityPackDetails.js
```diff
@@ -116,6 +117,7 @@ const renderAlerts = (pack) => {
 };
 
 const emptyStateContent = (pack, tabName) => {
+  const packUrl = pack.packUrl || REPO_URL;
```
zstix, line 120:
```suggestion
  const packUrl = pack.packUrl || QUICKSTARTS_REPO;
```

zstix (commented):
This looks great, just one minor comment about the name for the constant. I'll accept these changes and then ask the team for a re-review.

aswanson-nr (dismissed):

aswanson-nr (approved):

PR #1520, opened 2021-08-19
Update install button for doc-only quickstarts

### LINKED ISSUE #1513: [Public Catalog] Update the “Install” button to be “View installation docs” for integration packs
## Description
On the [ObservabilityPackDetails](https://github.com/newrelic/developer-website/blob/feature/o11y-packs/src/templates/ObservabilityPackDetails.js) template, the **Install** button on the **_integration packs_** should now read `View installation docs` and link to the documentation `url` referenced in the pack.

**!NOTE! The view repo button is not in this design, but it should be on this view.** 🤷 

integration_pack

`note:` an "integration" pack is not defined by type, so to decide which button to use, we'd need to look if the pack has install-able components (dashboards, alerts, etc) IF NONE, show view installation docs button

`note:` the pack migration will need to occur first. 

## Acceptance Criteria 

- [x] Ensure the CTA button on this details view is named `View installation docs`. 
- [x] On click the user should be directed to the correct documentation `url` that is referenced in the pack

zstix:
## Description
This updates the `<InstallButton />` component so that it will display "View installation docs" if there are no install-able components for the quickstart. In the event that there is no install-able components _and_ no documentation, the button will not be displayed (although we should never run into this scenario).

This also includes the PropType definition for a quickstart.

I started to go down the path of creating some tests for this component, but I ran into some difficulties mocking our theme 😕

## Related Issue(s)
* Closes #1513


src/types.js
```diff
@@ -12,3 +13,39 @@ export const link = PropTypes.shape({
...
+  name: PropTypes.string.isRequired,
+  url: PropTypes.string.isRequired,
+  description: PropTypes.string,
+  screenshots: PropTypes.arrayOf(PropTypes.string),
+});
+
+export const quickstartAlert = PropTypes.shape({
+  name: PropTypes.string.isRequired,
+  url: PropTypes.string.isRequired,
+  type: PropTypes.oneOf(['BASELINE', 'OUTLIER', 'STATIC']).isRequired,
```
aswanson-nr, line 27: We could break these types out into an enum, similar to the support level. Not required at all, just an idea.

aswanson-nr (approved):

src/types.js
zstix: Updated!

aswanson-nr (approved):

Issue #1516, opened 2021-08-16
[Public Catalog] Display the Documentation details for packs under the "Data source" tab


rudouglas:
## Description
Once the `Data source` tab is added to a pack as part of #1511 , if any new `documentation` info is available populate that data in the tab. This data should come from the `config.yml` e.g.:

```
documentation:
  - name: AWS RDS installation docs
    description: This is the description
    url: https://docs.newrelic.com/docs/integrations/amazon-integrations/aws-integrations-list/aws-rds-monitoring-integration
```

This should be a Tile component showing the 
 - `name` parameter as title of the tile
 - `description` field as other content in the tile
 - Link to the `url` when clicked. 

In future packs may have multiple references to documentation, so ensure that we account for this when displaying content in the tab.

**NOTE: the design image shows this tile under `Overview` tab, but it should be under `Data sources`**

<img width="419" alt="Screenshot 2021-08-16 at 14 07 20" src="https://user-images.githubusercontent.com/17122543/129568495-0f82fc1b-9489-46a5-a57f-3182b1a62ff6.png">

## Acceptance Criteria 

- [x] Populate the `Data sources` tab with the documentation reference defined in the pack.
- [x] Support multiple documentation links for packs with many doc references.

aswanson-nr:
@jpvajda We display a sort of "category" for each component at the bottom of the tile. 
<img width="363" alt="Screen Shot 2021-08-23 at 1 31 58 PM" src="https://user-images.githubusercontent.com/70179303/130515461-e0bca75d-7469-4955-9ce8-0cbfc1d3338a.png">
Do we want to add a "documentation" category to it?

zstix:
This has been merged into our feature branch. We're still missing some copy at the top of this tab, but we can tackle that when we receive that content (added to the production readiness list).

PR #1540, opened 2021-08-24
feat: add data sources tab content

aswanson-nr:
## Description

Adds content to the data sources tab 
NOTE: This is blocked until we have the `documentation` field in the API.

AC to get out of draft
- [x] Real documentation coming in and proofed
- [ ] Blurb for data sources tab is ready and added

* Closes https://github.com/newrelic/developer-website/issues/1516

### Light theme with tag (CURRENT)
Screen Shot 2021-08-25 at 3 36 06 PM

### Dark theme with tag (CURRENT)
Screen Shot 2021-08-25 at 3 35 51 PM

### Light theme with Icon (ALTERNATIVE OPTION)
Screen Shot 2021-08-24 at 3 53 38 PM

### Light theme with Icon (ALTERNATIVE OPTION)
Screen Shot 2021-08-24 at 3 53 48 PM


src/components/quickstarts/QuickstartDataSources.js
```diff
@@ -0,0 +1,77 @@
...
+          `}
+          interactive
+        >
+          <Link
+            to={doc.url}
+            css={css`
+              text-decoration: none;
+              color: inherit;
+            `}
+          >
```
zstix, line 49: If you want to avoid the extra element, you can provide an `as` prop to the `<Surface />` component so that it will be styled like the surface element, but have the functionality of `<Link />`:

```jsx
<Surface
  key={index}
  as={Link}
  base={Surface.BASE.PRIMARY}
  to={doc.url}
  css={/* ... */}
  interactive
>
```

src/components/quickstarts/QuickstartDataSources.js
```diff
@@ -0,0 +1,77 @@
...
+            <h3
+              css={css`
+                display: flex;
+                align-items: center;
+              `}
+            >
+              <Icon name="nr-documentation" />
```
zstix, line 56: I like the icon, but we are using the theme's `<Tag />` component to denote the classification for alert type and that pattern might make sense here too.

Either way, the reason this is fat is either the `stroke-width` or `font-weight`.

zstix (commented):
A couple of thoughts, but this is lookin' good.

nr-kkenney (approved):

PR #1547, opened 2021-08-26
chore: add dividing line and image slider refactor

rudouglas:
- Moved the Dots to below the Image slide so they are more visible
 - Dots do not show when there is only 1 image (same as left/right buttons)
 - Created a dividing line component to separate Image Sliders when a quickstart has more than 1


nr-kkenney:
looks good. i had one thing i saw, which we may keep separate so i still approved.

in the mobile view, the section headings and images are separated by variable blocks of whitespace, which looks a bit awkward.

Screen Shot 2021-08-27 at 4 52 16 PM

rudouglas:
@nr-kkenney ye this has to do with the height of the wrapping div being set, but the images being of different heights, I think we should put this in a nit/refactor of the image slider. I think a fix for this would require a design change tbh

nr-kkenney (approved):

PR #1566, opened 2021-08-30
Rename packId to quickstartId in install url

LizBaker:
This PR works in tandem with the refactor PR of the nerdlet. Changing the packId provided in the nr1 url to quickstartId in the dev site and consuming quickstartId in the nerdlet

### DIFF
--- a/src/data/constants.js
+++ b/src/data/constants.js
@@ -17,6 +17,8 @@ export const NR1_LOGIN_URL = 'https://staging-login.newrelic.com/login';
 // FIXME: update this to production URL when deployed / launched
 export const NR1_BASE_URL = 'https://staging-one.newrelic.com';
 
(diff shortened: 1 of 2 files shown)


nr-kkenney (approved):
Small question, could we define both a staging & production environment / url?

LizBaker:
@nr-kkenney the base url is going to be production once we get to a good point to come back and change it, and then you might use the base local one for development, but i don't think we need staging, does that make sense? or do you think they should be named differently?

nr-kkenney:
that makes sense. i was just curious why we would swap this later, when we could define all three and just swap between them -- maybe using something like an environment variable.

LizBaker:
@nr-kkenney totally down to do that when we go through and fix the todos!

PR #1568, opened 2021-08-31
Apply filter to images in dark mode

roadlittledawn:
## Description

Applies a filter to images in dark mode to avoid those bright white logo backgrounds. Attempts to not totally invert colors in most logos.

Not sure we actually want to do this, but it made my ailing eyes feel better ;)

## Screenshot(s)

https://user-images.githubusercontent.com/2952843/134957389-65b73845-bcb8-47e8-8d03-7fe1a8dfb688.mp4


jpvajda:
hmm not bad but let's see what @polfliet. What's tricky is we've changed colors on actual company and brand logos, which is not usually a good thing to do, so it could be confusing and also perhaps some partners could raise an issue.

jpvajda:
We talked about aligning a little more closely with the darkmode filter in the product so we can do a better job displaying red and blue tones. Once we make that improvement this can be production ready @roadlittledawn

nr-kkenney (approved):

Issue #1559, opened 2021-08-27
[Public Catalog] Use different layout for quickstart details pages


roadlittledawn:
## Summary

We want to change the layout of a quick start details page so:
* there is no left sidebar, and content takes up full width of page
* include breadcrumbs at top (similar to NR1 catalog details page)
* exclude mobile nav hamburger menu

Consider creating `Breadcrumbs` component in the theme for docs site to use in the future.

## Remaining work in branch

Some of the work here is already in [this branch](https://github.com/newrelic/developer-website/tree/clang/quick-start-layout). Remaining work includes:
* include breadcrumbs at top (similar to NR1 catalog details page)

### Mock

![2021-08-27_13-58-24](https://user-images.githubusercontent.com/2952843/131188044-d1abe1c8-8827-405b-abad-f641d69c47f6.png)

**breadcrumbs in NR1**
![2021-08-30_11-48-36](https://user-images.githubusercontent.com/2952843/131389407-a32cbdc8-4fe6-4794-884f-0ee14ec1c174.png)

## Acceptance criteria
- [x] Each quick start page is full page width (like mock), no sidebar
- [x] If I view details page on mobile, there's no hamburger menu / nav and page is responsive
- [x] As a user, I can get a sense of where I am by viewing breadcrumbs, and clicking on link to quick starts landing

## Related links
https://github.com/newrelic/developer-website/issues/1534
https://github.com/newrelic/developer-website/tree/clang/quick-start-layout

nr-kkenney:
closing since all prs merged, criteria checked off.

PR #1574, opened 2021-08-31
Add breadcrumbs to details pages

roadlittledawn:
## Description

Adds a `Breadcrumbs` component and uses it on the pack details pages above the page title / pack name.

Stopped short of adding to theme for time being, but made it its own component so it's a wee bit easier to port over if/when we want to.

Closes https://github.com/newrelic/developer-website/issues/1559

### DIFF
--- a/src/components/Breadcrumbs.js
+++ b/src/components/Breadcrumbs.js
@@ -8,12 +8,6 @@ const Breadcrumbs = ({ segments, separator }) => {
     <div
       css={css`
         margin: 2em 0;
-        > :first-child {
-          margin-left: 0;
-        }
-        > * {
-          margin-left: 0.5em;
-        }
       `}
       aria-label="breadcrumb"
     >
@@ -27,13 +21,11 @@ const Breadcrumbs = ({ segments, separator }) => {
           <span
             key={`breadcrumb-${segment.name}`}
             css={css`
-              :after {
-                margin-left: 0.5em;
-                content: ' ${separator} ';
-              }
-              :last-child {
+              :not(:last-of-type) {
                 :after {
-                  content: '';
+                  margin: 0 0.5em;
+                  display: inline-block;
+                  content: '${separator}';
                 }
               }
             `}
--- a/src/components/Breadcrumbs.js
+++ b/src/components/Breadcrumbs.js
@@ -1,0 +1,52 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { css } from '@emotion/react';
+import { Link } from '@newrelic/gatsby-theme-newrelic';
+
+const Breadcrumbs = ({ segments, separator }) => {
+  return (
+    <div
+      css={css`
+        margin: 2em 0;
+        > :first-child {
+          margin-left: 0;
+        }
+        > * {
+          margin-left: 0.5em;
+        }
+      `}
+      aria-label="breadcrumb"
+    >
+      {segments
+        .map((segment) => {
+          const elem = segment.url ? (
+            <Link to={segment.url}>{segment.name}</Link>
+          ) : (
+            segment.name
+          );
+          return <span key={`breadcrumb-${segment.name}`}>{elem}</span>;
+        })
+        .reduce((prev, curr) => [
+          prev,
+          React.createElement('span', {}, separator),
+          curr,
+        ])}
+    </div>
+  );
+};
+
+Breadcrumbs.defaultProps = {
+  separator: '/',
+};
+
+Breadcrumbs.propTypes = {
+  segments: PropTypes.arrayOf(
+    PropTypes.shape({
+      name: PropTypes.string.isRequired,
+      url: PropTypes.string,
+    })
+  ),
+  separator: PropTypes.string,
+};
+
+export default Breadcrumbs;
--- a/src/components/Breadcrumbs.js
+++ b/src/components/Breadcrumbs.js
@@ -56,7 +56,7 @@ Breadcrumbs.propTypes = {
       name: PropTypes.string.isRequired,
       url: PropTypes.string,
     })
-  ),
+  ).isRequired,
   separator: PropTypes.string,
 };
 
--- a/src/components/Breadcrumbs.js
+++ b/src/components/Breadcrumbs.js
@@ -17,20 +17,31 @@ const Breadcrumbs = ({ segments, separator }) => {
       `}
       aria-label="breadcrumb"
     >
-      {segments
-        .map((segment) => {
-          const elem = segment.url ? (
-            <Link to={segment.url}>{segment.name}</Link>
-          ) : (
-            segment.name
-          );
-          return <span key={`breadcrumb-${segment.name}`}>{elem}</span>;
-        })
-        .reduce((prev, curr) => [
-          prev,
-          React.createElement('span', {}, separator),
-          curr,
-        ])}
+      {segments.map((segment) => {
+        const elem = segment.url ? (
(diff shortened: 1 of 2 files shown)


src/components/Breadcrumbs.js
```diff
@@ -0,0 +1,52 @@
...
+          const elem = segment.url ? (
+            <Link to={segment.url}>{segment.name}</Link>
+          ) : (
+            segment.name
+          );
+          return <span key={`breadcrumb-${segment.name}`}>{elem}</span>;
+        })
+        .reduce((prev, curr) => [
+          prev,
+          React.createElement('span', {}, separator),
```
zstix, line 31: One option, if we wanted to avoid creating a new react component and HTML element, is to use emotion and the `:after` pseudo-element:

```jsx
segments.map((segment) => (
  <Link
    css={css`
      :after {
        content: "${separator}";
      }
    `}
    /* other props / JSX */
))
```

src/components/Breadcrumbs.js
```diff
@@ -0,0 +1,52 @@
...
+      aria-label="breadcrumb"
+    >
+      {segments
+        .map((segment) => {
+          const elem = segment.url ? (
+            <Link to={segment.url}>{segment.name}</Link>
+          ) : (
+            segment.name
+          );
+          return <span key={`breadcrumb-${segment.name}`}>{elem}</span>;
```
zstix, line 27: This _could_ be simplified to something like this, if we wanted to avoid wrapping the link in an extra `<span>` tag:

```jsx
return segment.url
  ? <Link key={segment.name} to={segment.url}>segment.name</Link>
  : <span key={segment.name}>{segment.name}</span>;
```

src/components/Breadcrumbs.js
```diff
@@ -0,0 +1,52 @@
...
+Breadcrumbs.defaultProps = {
+  separator: '/',
+};
+
+Breadcrumbs.propTypes = {
+  segments: PropTypes.arrayOf(
+    PropTypes.shape({
+      name: PropTypes.string.isRequired,
+      url: PropTypes.string,
+    })
```
zstix, line 47:
```suggestion
    }).isRequired
```

src/components/Breadcrumbs.js
zstix: If you wanted to go the element-choosing-route, you could also consider something like this:

```jsx
const Component = segment.url ? Link : span;
return <Component key={segment.name} to={segment.url}>{segment.name}</Component>
```

zstix (commented):
This is lookin' great! I had a few comments of various other ways we could approach some of this stuff.

src/components/Breadcrumbs.js
roadlittledawn: ah yes, good ol' psuedo selector. didn't think of that. i like it!

src/components/Breadcrumbs.js
roadlittledawn: now that i'm adding the separator via `:after` psuedo selector, i can't add that to the link because it becomes part of the link, so i need it inside a generic element like `span` i think.

src/components/Breadcrumbs.js
```diff
@@ -0,0 +1,63 @@
...
+            css={css`
+              :after {
+                margin-left: 0.5em;
+                content: ' ${separator} ';
+              }
+              :last-child {
+                :after {
+                  content: '';
+                }
+              }
```
zstix, line 38: There was [a great article](https://css-tricks.com/you-want-enabling-css-selectors-not-disabling-ones/) (and linked articles) published about this topic yesterday. The author proposes that we should strive for "enabling" selectors rather than creating a generic selector and then an additional one to "unset" a style.

Here's how the author would recommend this block. I also added some rules to prevent you from needing to add some whitespace _characters_ on either side of the separator.

```css
:not(:last-of-type) {
  margin-left: 0.5em;
  padding: 0 0.5em;
  display: inline-block;
  content: ${separator};
}
```

src/components/Breadcrumbs.js
```diff
@@ -0,0 +1,63 @@
...
+        }
+        > * {
+          margin-left: 0.5em;
+        }
+      `}
+      aria-label="breadcrumb"
+    >
+      {segments.map((segment) => {
+        const elem = segment.url ? (
+          <Link to={segment.url}>{segment.name}</Link>
```
zstix, line 22: One thing I thought of after reviewing this yesterday: should we _always_ present the last breadcrumb segment as a `<span>` rather than a link? There's not much sense in linking to the page the user is on and, if the component handles that, we wont need to do weird stuff like `url: null` (like we're doing in `QuickstartDetails.js`)

src/templates/QuickstartDetails.js
```diff
@@ -49,6 +50,16 @@ const QuickstartDetails = ({ data, location }) => {
...
   const packUrl = pack.packUrl || QUICKSTARTS_REPO;
   const tessen = useTessen();
+  const breadcrumbSegments = [
+    {
+      name: 'Instant Observability (I/O)',
+      url: '/instant-observability/',
+    },
+    {
+      name: pack.name,
+      url: null,
```
zstix, line 60: I left a comment about this in the other file. Even if we decide we _don't_ want to account for this in the component, I think it might be better to just not supply this value. `undefined` and `null` will both evaluate to `false` and I think it's a little cleaner this way:

```suggestion
```

src/templates/QuickstartDetails.js
```diff
@@ -49,6 +50,16 @@ const QuickstartDetails = ({ data, location }) => {
   const pack = data.quickstarts;
   const packUrl = pack.packUrl || QUICKSTARTS_REPO;
   const tessen = useTessen();
+  const breadcrumbSegments = [
```
zstix, line 53: Totally a style thing (zero pressure to change): I'm curious why you opted for this name?

***breadcrumb***Segments / breadcrumb***Segments*** 

It seems like it has a higher level of specificity than we need. Are there _other_ types of segments? Are there segments for _other_ things? We might be able to get away with less verbosity: something like `breadcrumbs` might be easier to understand.

Or perhaps we don't want to misrepresent what the variable represents! I could be convinced either way...just pondering 🤔

zstix (commented):
Sorry...a couple more thoughts. Most of these are more of me pondering because I haven't had enough caffeine.

src/components/Breadcrumbs.js
roadlittledawn: oh nice. yeah, when i wrote that i thought it was a little weird / funky to have to undo it for the `:last-child`. added that with slight tweak

src/components/Breadcrumbs.js
roadlittledawn: yeah i thought about that. i know sometimes tech docs folks liked having things like breadcrumbs be links so they could copy/paste the link. and when i looked at google's material design for inspiration they used a link for the last segment sometimes, so thought maybe there's something to it and left it as optionally available.

src/templates/QuickstartDetails.js
roadlittledawn: 🤔 no strong opinion here, other than i do like the shorter `breadcrumbs`. made that change.

src/templates/QuickstartDetails.js
roadlittledawn: oh right. thanks!

zstix (approved):
🍞  🥖  🍞  🥖  🍞

PR #1580, opened 2021-09-01
Fix `logger` API page

zstix:
## Description
This PR adds a few fixes to reduce the number of errors on the developer site. More specifically, this removes:

* The warning that we needed a unique `key` for the `<meta>` tags coming from our SEO component
* The warning coming from the `logger` API page (due to the SDK being in a weird state - see screenshots)

## Screenshot(s)
It looks like the SDK for the `logger` API is mostly incomplete. A number of the adjustments in the PR were defensive adjustments to get the page to render without errors. That said, it's still pretty incomplete:

Screen Shot 2021-09-01 at 16 19 31

### DIFF
--- a/src/components/FunctionDefinition.js
+++ b/src/components/FunctionDefinition.js
@@ -4,7 +4,12 @@ import CodeDef from './CodeDef';
 import { graphql } from 'gatsby';
 
 const FunctionDefinition = ({ className, arguments: params, returnValue }) => {
-  if (!params.length && !returnValue && !returnValue.length) {
+  const hasParams = params.length;
+  const isReturnArray = Array.isArray(returnValue);
+  const hasReturn =
+    (isReturnArray && returnValue.length) || (!isReturnArray && returnValue);
+
+  if (!hasParams && !hasReturn) {
     return null;
   }
 
--- a/src/components/FunctionDefinition.js
+++ b/src/components/FunctionDefinition.js
@@ -4,8 +4,12 @@ import CodeDef from './CodeDef';
 import { graphql } from 'gatsby';
 
 const FunctionDefinition = ({ className, arguments: params, returnValue }) => {
-  if (!params.length && !returnValue.length) {
+  if (!params.length && !returnValue && !returnValue.length) {
     return null;
+  }
+
+  if (!Array.isArray(returnValue)) {
+    returnValue = [returnValue];
   }
 
   return (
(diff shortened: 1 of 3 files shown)


src/components/FunctionDefinition.js
```diff
@@ -4,10 +4,14 @@ import CodeDef from './CodeDef';
 import { graphql } from 'gatsby';
 
 const FunctionDefinition = ({ className, arguments: params, returnValue }) => {
-  if (!params.length && !returnValue.length) {
+  if (!params.length && !returnValue && !returnValue.length) {
```
nr-kkenney, line 7: i think this line is evaluating differently than before, but im not sure if thats intended.

in particular, i think `!returnValue` will return false on empty array of return value. but i also see how now if passed something that isn't array, thats accounted for.

zstix: Oh good catch!

I think we'll want to ensure the following:
* Are there no parameters AND:
  * Is the returnValue an array & is it empty OR:
    * Is it not an array and falsey

If that's correct, then here's what I've got:

```suggestion
  const hasParams = params.length;
  const isReturnArray = Array.isArray(returnValue);
  const hasReturn = (isReturnArray && returnValue.length) || (!isReturnArray && returnValue);
  
  if (!hasParams && !hasReturn) {
```

Does that seem right to you?

nr-kkenney: i think this is right. would you happen to have any thoughts on splitting this into multiple if checks, each with a return for one condition.

zstix: hmmm, let's see how that would look:

```suggestion
if (!params.length) return;
if (Array.isArray(returnValue) && !returnValue.length) return;
if (!Array.isArray(returnValue) && !returnValue) return;
```

That might be a little simpler (and could be combined with the boolean variables for maximum readability). Ideally we don't run into this situation that often - this code is mostly just extra defense against the SDK being in the wrong format.

Aaaanyway, what do you think?

nr-kkenney: my thought was just around all the boolean logic in one line, and trying to think through it. i defer to you for best practice / experience 🙈  i would approve either.

zstix: Alright - I'll probably go with the first one (one-line `if` statements make me nervous). Thanks for helping me work through that!

nr-kkenney (approved):

PR #1573, opened 2021-08-31
chore: updating relicans and nerdlog pages

laurenelee:
## Description

added Lauren Lee to nerdlog and relicans and removed prior teammates 

## Reviewer Notes

If there are links or steps needed to test this work, add them here.

## Related Issue(s) / Ticket(s)

If there are any related [GitHub Issues](https://github.com/newrelic/developer-website/issues) or JIRA tickets, add links to them here.
* [issue]()

## Screenshot(s)

If relevant, add screenshots here.

## Use Conventional Commits

Please help the maintainers by leveraging the following [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/)
standards in your pull request title and commit messages.

## Use `chore`

* for minor changes / additions / corrections to content.
* for minor changes / additions / corrections to images.
* for minor non-functional changes / additions to github actions, github templates, package or config updates, etc

```bash
git commit -m "chore: adjusting config and content"
```

## Use `fix`

* for minor functional corrections to code.

```bash
git commit -m "fix: typo and prop error in the code of conduct"
```

## Use `feat`

* for major functional changes or additions to code.

```bash
git commit -m "feat(media): creating a video landing page"
```

PR #1581, opened 2021-09-02
Relicans update

jpvajda:
Resolves yarn lock issues for https://github.com/newrelic/developer-website/pull/1573 and merged in the relican page changes.


jpvajda:
helping @laurenelee to get this PR to pass checks, I'm curious as why there is still a yarn lock file conflict...

jpvajda:
GOT IT.. 🥳 🎈

LizBaker:
I noticed Ali's last name was updated in the relican's page but not the nerdlog page, assuming it should change in both but just want to double check

jpvajda:
@LizBaker good catch I change the name in `nerdlogData.js` which is where the name was sourced from on that page.

nr-kkenney (approved):
lgtm

Issue #1539, opened 2021-08-24
[Swiftype] Add meta tags to details pages


roadlittledawn:
## Summary

Add the following `<meta>` tags to the `<head>` of all quick start detail pages so we can use that in swiftype search queries.

The theme's [SEO component](https://github.com/newrelic/gatsby-theme-newrelic/blob/develop/packages/gatsby-theme-newrelic/src/components/SEO.js) does this automatically by looking at the domain of a given site's `siteUrl` (set in `gatsby-config.js`). We will need to include a way to pass the `type` for quick starts (because they live on developer subdomain) as a prop to override this default behavior. 

```
<meta class="swiftype" name="type" data-type="enum" content="io">
```

Add a `<meta>` tag for each keyword associated with a quick start:

```
<meta class="swiftype" name="tags" data-type="string" content="TAG_NAME_1">
<meta class="swiftype" name="tags" data-type="string" content="TAG_NAME_2">
```

Add a `<meta>` tag for the quick start name:

```
<meta class="swiftype" name="quick_start_name" data-type="string" content="QUICK_START_NAME">
```
the content should populate with `quickstart.title` field.

## Notes

In order to see those meta tags in development, I _think_ you may have to run a build.

## Acceptance criteria
- [ ] On each quick start details page, the above meta tags are in the `<head>` element
- [ ] On each quick start page, the only `type` field is `quick_starts` (and not `developer`)

## Relavant link
[Theme's SEO component](https://github.com/newrelic/gatsby-theme-newrelic/blob/develop/packages/gatsby-theme-newrelic/src/components/SEO.js)

LizBaker:
@roadlittledawn i see the site type here as both "io" and "quick_starts" and somehow my change ended up being "quickstarts" 😂 which one of these would you prefer?

roadlittledawn:
my bad. didn't get a chance to update all the tickets to use same value. i believe we said we'll go with `quickstarts`

PR #1585, opened 2021-09-02
Add type and qs name to meta tags

LizBaker:
This adds a `quickstart` type to the quickstart landing page and the details view, which will override the `developer` type. It also adds a 'quick_start_name' meta tag

Partially addresses https://github.com/newrelic/developer-website/issues/1539

To review:
- [x] In the site preview inspect a normal dev site page and make sure the meta tag for type has 'developer' as the content (you can command f for swiftype)
- [x] On the catalog page make sure the meta tag for type has 'quickstarts' as the content
- [x] On a details page make sure there is a new tag called 'quick_start_name' with the name as it's contents **this is using quickstart.title and not all quickstarts have titles (like apache) so be sure to check one that does


rudouglas (commented):
I don't see the `quick_start_name` tag in any of the packs that have titles, but I may be missing something. For example, [aws-ebs](https://github.com/newrelic/newrelic-observability-packs/blob/4185d56c5f2028f6ac96f4d1fe67624d2d256f69/packs/aws/aws-ebs/config.yml) has a `title` but there is no `quick_start_name` tag in the page source. 

There is this tag labelled `title` but not sure if that's related:
```
<meta data-react-helmet="true" name="title" class="swiftype" data-type="string" content="aws-ebs"/>
```

LizBaker:
@rudouglas looks like we have no titles in our quickstarts.json, hasn't been updated in awhile! i'll add that in since we should update anyway

src/templates/QuickstartDetails.js
```diff
@@ -75,7 +75,12 @@ const QuickstartDetails = ({ data, location }) => {
 
   return (
     <>
-      <DevSiteSeo title={pack.name} location={location} />
+      <DevSiteSeo
+        title={pack.name}
+        type="quickstarts"
+        location={location}
+        quickStartName={pack.title}
```
zstix, line 82: It looks like the `<DevSiteSeo>` component [accepts a `meta` prop](https://github.com/newrelic/developer-website/blob/a9b39ff31eb7add2205f28d9dc680dfd596ff2cd/src/components/DevSiteSeo.js#L121) for supplying custom meta tags as needed. Perhaps we should take advantage of that, rather than extending the component for just this one use-case.

If we did that, we'd build the meta tags for the quickstart in this file and just pass them in like so:

```jsx
const quickstartMeta = [
  {
    name: 'quick_start_name',
    class: 'swiftype',
    'data-type': 'string',
    content: pack.title,
  },
  {
    name: 'type',
    class: 'swiftype',
    'data-type': 'enum',
    content: 'quickstarts',
  },
];

// ...

<DevSiteSeo
  title={pack.title}
  type="quickstarts"
  location={location}
  meta={quickstartMeta}
/>
```

zstix (commented):
One suggestion regarding how we use the `<DevSiteSeo>` component.

src/templates/QuickstartDetails.js
LizBaker: works for me! 🎉

zstix (approved):

PR #1586, opened 2021-09-02
Alec/landing layout

aswanson-nr:
## Description

Various updates to the IO marketplace. 
* Refactor search/filtering to reduce LoC and some complexity

- [x] Refactor search/filter
- [x] add tessen and page actions

PR #1591, opened 2021-09-04
Filter and search quickstarts with hooks

zstix:
## Description
This pull request introduces a new hook called `useFilterSearchSort` that handles keeping track of arbitrarily complex lists that can (optionally) be filtered, searched, and sorted. The goal of this hook is to abstract away the logic needed to update the URL and track events and allows components to just focus on rendering things.

What this does:
* Keeps track of search, filter, and sort state
* Updates URL parameters (if `updateURL` is set to `true`)
* Instrument with New Relic Browser (if a `trackingPrefix` is supplied)
* Instrument with Tessen (if `trackingPrefix` and `trackingCategory` are supplied)
* Debounces URL updates and instrumentation when a user is searching

What this does _not_:
* Keep track of the items being sorted
* Update the _list_ of filters (only tracks the actively selected filter(s))
* Making the state available via context (a decent idea, I just haven't implemented it yet, see the `@todo`)

Due to it's generalized format, it should be fairly straightforward to bring this into the theme and use it on other pages. For example, here's a condensed view of how we're using it on the _Instant Observability_ page (filter in 1 way, include a search, instrument it):

```jsx
const { search, setSearch, filters, setFilters } = useFilterSearchSort({
  filters: { type: 'all' },
  trackingCategory: 'InstantObservability',
  trackingPrefix: 'quickstart',
  updateURL: true
});

// ...

<input onChange={(e) => setSearch(e.target.value)} value={search}>
<select onChange={(e) => setFilters({ ...filters, type: e.target.value })}>
```

And here's how we _could_ use it for the [icon gallery](https://developer.newrelic.com/components/icon/) on the developer site (just search, no instrumentation):

```jsx
const { search, setSearch } = useFilterSearchSort();

// ...

<input onChange={(e) => setSearch(e.target.value)} value={search} />
```

And here's how we _could_ use it for the [attribute dictionary](https://docs.newrelic.com/attribute-dictionary/) on the docs website (filter in 2 ways, include a search, instrument it):

```jsx
const { search, setSearch, filters, setFilters } = useFilterSearchSort({
  filters: {
    dataSource: 'all',
    event: 'all',
  },
  trackingCategory: 'AttributeDictionary',
  trackingPrefix: 'attributeDictionary',
  updateURL: true
});

// ...

<select onChange={(e) => setFilters({ ...filters, dataSource: e.target.value })}>
<select onChange={(e) => setFilters({ ...filters, event: e.target.value })}>
```

This reduces the amount of boilerplate code we have to write when implementing these types of user interfaces. This also makes it fairly trivial to add new functionality or instrumentation.


zstix:
Closing this in favor of #1586

PR #1594, opened 2021-09-07
Adds a banner component for the IO catalog

LizBaker:
The code for this PR is ready for review, however, it will need an updated image and video link before merging

- [x] placeholder replaced with actual image
- [ ] ~correct video link~ separate ticket https://github.com/newrelic/developer-website/issues/1616

Screen Shot 2021-09-13 at 3 18 25 PM
Screen Shot 2021-09-13 at 3 17 50 PM


nr-kkenney (approved):

src/components/IOBanner.js
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import { css } from '@emotion/react';
+import useMobileDetect from 'use-mobile-detect-hook';
+import { Button, Link } from '@newrelic/gatsby-theme-newrelic';
+import ioBanner from '../images/ioBanner.svg';
+
+const IOBanner = () => {
+  const detectMobile = useMobileDetect();
```
aswanson-nr, line 8: This could be moved into a prop since the instant-observability page is showing/hiding based on the same hook! Up to you though

aswanson-nr (dismissed):
Code looks great overall! I noticed same styling breakage at certain breakpoints and attached a screenshot 
Screen Shot 2021-09-13 at 3 23 01 PM

My other comment is just an opinion and can be ignored!

src/components/IOBanner.js
LizBaker: true! it's kinda nice if it can take care of itself if it got used elsewhere (though what are the chances 😂) but a prop seems good in this situation

zstix (commented):
This looks great! Can we remove `src/images/ioBanner.png` since we now have a SVG banner?

src/components/IOBanner.js
```diff
@@ -0,0 +1,83 @@
...
+          css={css`
+            color: var(--color-white);
+            font-size: ${isMobile ? '1rem' : '1.5vw'};
+          `}
+        >
+          Instant observability
+        </h1>
+        <p
+          css={css`
+            font-size: 0.75vw;
```
zstix, line 48: I noticed that this scales the font super small at certain browser widths. Perhaps we can use [`max()`](https://developer.mozilla.org/en-US/docs/Web/CSS/max()):

```suggestion
            font-size: max(10pt, 0.75vw);
```

If we like that, we might want to update the heading too.

LizBaker: ah yes, that's handy

zstix (approved):

PR #1596, opened 2021-09-08
Remove /nerdlet from nr1 url

LizBaker:
We were adding `/nerdlet/catalog-pack-details.catalog-pack-contents` to the nr1 url when installing a pack. i don't think it's possible to load a nerdlet this way and it seems it redirects for some users and not others. This removes that part of the url (but maintains the nerdlet ID passed in the pane param) which uses the platform to launch our nerdlet. if we don't want to see the platform tabs at the top (explorer, alerts, etc) we can make our own launcher, but in the meantime this will at least render things functional


nr-kkenney (approved):
still works for me 👍

PR #1608, opened 2021-09-10
Ru/featured filter

rudouglas:
Screenshot 2021-09-10 at 15 53 28

Screenshot 2021-09-10 at 16 15 37

### DIFF
--- a/src/components/PackTile.js
+++ b/src/components/PackTile.js
@@ -8,6 +8,7 @@ import {
   Icon,
   useTessen,
   useInstrumentedHandler,
+  Tag,
 } from '@newrelic/gatsby-theme-newrelic';
 import PackImg from './PackImg';
 
@@ -23,6 +24,7 @@ const VIEWS = {
 const PackTile = ({
   id,
   view,
+  featured,
   name,
   fields,
   logoUrl,
@@ -69,11 +71,13 @@ const PackTile = ({
       interactive
       css={css`
         overflow: hidden;
+        display: flex;
+        flex-direction: column;
 
         ${view === VIEWS.LIST &&
         css`
-          display: flex;
           margin-bottom: 1em;
+          flex-direction: row;
         `}
       `}
       onClick={fields ? handlePackClick : handleBuildTileClick}
@@ -93,6 +97,7 @@ const PackTile = ({
           css`
             max-height: 150px;
 
+            flex: 0 0 auto;
             @media (max-width: 1080px) {
               display: none;
             }
@@ -102,11 +107,11 @@ const PackTile = ({
       <div
         css={css`
           padding: 1em;
-
+          flex: 1 1 auto;
           ${view === VIEWS.LIST &&
           css`
-            width: 75%;
-
+            width: 100%;
+            flex: 1 1 auto;
             @media (max-width: 1080px) {
               width: 100%;
             }
@@ -125,6 +130,31 @@ const PackTile = ({
           {summary}
         </p>
       </div>
+      <div
+        css={css`
+          padding: 1em;
+          display: flex;
+          justify-content: flex-end;
+          ${view === VIEWS.LIST &&
+          css`
+            flex-direction: column;
+            justify-content: flex-end;
+            @media (max-width: 1080px) {
+              width: 100%;
+            }
+          `}
+        `}
+      >
+        {featured && (
+          <Tag
+            css={css`
+              background-color: var(--color-brand-300);
+            `}
+          >
+            Featured
+          </Tag>
+        )}
+      </div>
     </Surface>
   );
 };
@@ -140,6 +170,7 @@ PackTile.propTypes = {
   summary: PropTypes.string,
   level: PropTypes.string,
   className: PropTypes.string,
+  featured: PropTypes.bool,
 };
 
 export default PackTile;
--- a/src/components/PackTile.js
+++ b/src/components/PackTile.js
@@ -91,7 +91,7 @@ const PackTile = ({
           object-fit: scale-down;
           width: ${view === VIEWS.GRID ? 100 : 25}%;
           padding: 0 ${view === VIEWS.GRID ? 5 : 1}%;
-          margin: ${view === VIEWS.GRID ? 'auto' : 0};
+          margin: 10px auto 10px auto;
 
           ${view === VIEWS.LIST &&
           css`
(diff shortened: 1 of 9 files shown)


src/data/quickstarts.json
```diff
@@ -3118,6 +3120,7 @@
         "url": "https://docs.newrelic.com/docs/integrations/open-source-telemetry-integrations/open-source-telemetry-integration-list/new-relics-dropwizard-integration"
       }
     ],
+    "keywords": ["featured"],
```
jpvajda, line 3123: Did we manually set this to test? or are these actually set at the QS level now?

src/data/quickstarts.json
```diff
@@ -3391,6 +3394,7 @@
         "url": "https://raw.githubusercontent.com/newrelic/newrelic-observability-packs/v0.12.0/packs/gatsby-build/dashboards/gatsby-build.json"
       }
     ],
+    "keywords": ["featured"],
```
jpvajda, line 3397: What occurs when we actually merge in the quick start data and the actual keywords are set, will these be removed? I'm asking as these aren't actual featured QSs, so just want to make sure this is just for testing reasons.

jpvajda (changes requested):

src/components/PackTile.js
```diff
@@ -23,6 +24,7 @@ const VIEWS = {
 const PackTile = ({
   id,
   view,
+  featured,
```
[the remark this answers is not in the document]
jpvajda: Confirmed Featured should be at the top.

jpvajda (dismissed):

src/data/quickstarts.json
rudouglas: I manually set this to test for the moment

src/data/quickstarts.json
rudouglas: Ye when we manually update with proper keywords, any test keywords i've entered will be removed so no stress

src/components/PackTile.js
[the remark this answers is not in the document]
rudouglas: oh yes that would make sense 🥇

src/components/PackTile.js
```diff
@@ -87,12 +91,13 @@ const PackTile = ({
           object-fit: scale-down;
           width: ${view === VIEWS.GRID ? 100 : 25}%;
           padding: 0 ${view === VIEWS.GRID ? 5 : 1}%;
-          margin: ${view === VIEWS.GRID ? 'auto' : 0};
+          margin: 10px auto 10px auto;
```
zstix, line 94: This could be simplified to:
```suggestion
          margin: 10px auto;
```

src/pages/instant-observability.js
```diff
@@ -18,10 +18,12 @@ import { navigate } from '@reach/router';
 
 import BUILD_YOUR_OWN from '../images/build-your-own.svg';
 import { useDebounce } from 'react-use';
+import { sortFeaturedPacks } from '../utils/sortFeaturedPacks';
 
 const { QUICKSTARTS_REPO } = require('../data/constants');
 
 const FILTERS = [
+  { name: 'Featured', type: 'featured', icon: 'nr-relicans' },
```
zstix, line 26: I _think_ we probably want to handle keyword filters differently. This array (which could probably be renamed) is for filtering by _content_.

If we add `Featured` to this array, it means we wont be able to filter by featured quickstarts _and_ ones that include dashboards. We should probably have a separate list of keyword filters with their own function(s) to narrow down the list. I'm sure @aswanson-nr has some thoughts here too.

src/utils/sortFeaturedPacks.js
```diff
@@ -0,0 +1,9 @@
+export const sortFeaturedPacks = (packs) => {
+  const featuredPacks = packs.filter((pack) =>
+    pack.keywords?.includes('featured')
+  );
+  const nonFeaturedPacks = packs.filter(
+    (pack) => !pack.keywords?.includes('featured')
+  );
+  return [...featuredPacks, ...nonFeaturedPacks];
+};
```
zstix, line 9: I think we can probably use [`Array.prototype.sort()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) to handle this, rather than doing two passes through the list. `O(n)` is more performant than `O(2n)`. Here's how that could look:

```js
const isFeatured = (quickstart) =>
  quickstart.keywords && quickstart.keywords.includes('featured');

const sortedQuickstarts = quickstarts.sort((a, b) => {
  if (isFeatured(a) && !isFeatured(b)) {
    return 1;
  }
  if (isFeatured(b) && !isFeatured(a)) {
    return -1;
  }
  return 0;
});
```

In general, I like to favor built-in methods that allow the code to be a little more declarative about the operation we're trying to perform. We're trying to sort a list, so we use the `sort` method.

Also, I'm not sure if we have access to optional chaining (e.g. `foo?.bar?.baz`).

Also Also, we should really avoid references to "packs". This will help when we go to production readiness and need to update every instance to be "quickstarts".

zstix (changes requested):
I had a couple of comments that we might want to address before merging this in.

src/utils/sortFeaturedPacks.js
rudouglas: ahhh yes that makes more sense, i thought I couldn't use a sort but I did not think about using `isFeatured` as a boolean :aha:

src/pages/instant-observability.js
```diff
@@ -110,11 +128,12 @@ const QuickstartsPage = ({ data, location }) => {
...
   const quickstarts = data.allQuickstarts.nodes;
 
-  const sortedQuickstarts = sortFeaturedPacks(quickstarts);
+  const sortedQuickstarts = sortFeaturedQuickstarts(quickstarts);
 
   const filteredQuickstarts = sortedQuickstarts
     .filter(filterBySearch(search))
-    .filter(filterByContentType(filter));
+    .filter(filterByContentType(filter))
+    .filter(filterByContentType(category));
```
aswanson-nr, line 136: To keep things simple, let's break this filter out into a keywords/category function. That way the `filterByContentType` function doesn't have extra opaque functionality

rudouglas: Sorry this function was designed to sort either content but it makes sense to split it out so it's easier to understand

aswanson-nr (approved):
The featured filter cannot be deselected, but since we are making changes to that section in the coming sprint I think it's fine to merge as is.

PR #1613, opened 2021-09-13
chore: move CTA buttons from page tools

rudouglas:
Screenshot 2021-09-13 at 14 11 20
Screenshot 2021-09-13 at 14 12 35


zstix (commented):
Looks good to me. Not sure why the lockfile got updated, but I didn't run into any issue with it on my machine.

zstix (approved):

PR #1620, opened 2021-09-14
chore: replace name with title where applicable

rudouglas:
I didn't change the tessen events because i i think the name makes more sense for searchability, plus it aligns with what we see in the url

I also notice the EmptyTab wasn't rendering correctly so I fixed that


src/components/PackTile.js
```diff
@@ -80,7 +81,7 @@ const PackTile = ({
     >
       <PackImg
         logoUrl={logoUrl}
-        packName={name}
+        packName={title || name}
```
LizBaker, line 84: i think we probably don't need the `or` here if we know title will be required (and possibly name is going away?) lemme know if that's wrong though
```suggestion
        packName={title}
```

LizBaker (approved):
One little nit but otherwise 🐑 it!

jpvajda:
I picked this up for @rudouglas and resolved a conflict in the packTile.js file this should be good to go now

jpvajda:
actually even though the gatsby build checks pass, it's loading a blank page

https://build-9515bdfc-f914-40f3-a9fb-e6a464e61f50.gtsb.io/instant-observability/

jpvajda:
I added back the OR statement and it's seem to render OK, but now some packs seem to have lost their summaries... 

Screen Shot 2021-09-15 at 5 31 43 PM

Screen Shot 2021-09-15 at 5 31 54 PM

LizBaker (dismissed):
✨

LizBaker:
just to document: we realized some packs are still missing summaries and those will need to be added. The missing titles are also expected until all packs are updated with a title

aswanson-nr:
Looks like the "Build you own pack" needs to have it's title updated, I'll add a commit. @LizBaker I'm going to dismiss your's

src/pages/instant-observability.js
```diff
@@ -445,7 +445,7 @@ const QuickstartsPage = ({ data, location }) => {
                 }
                 view={view}
                 logoUrl={BUILD_YOUR_OWN}
-                name="Build your own quickstart"
+                title="Build your own quickstart"
                 summary="Can't find a pack with what you need? Check out our README and build your own."
```
LizBaker, line 449:
```suggestion
                summary="Can't find a quickstart with what you need? Check out our README and build your own."
```

LizBaker (approved):

Issue #1603, opened 2021-09-09
Final instrumentation checks


aswanson-nr:
## Description
Most of the instrumentation has been completed as we worked through the project, this is just to catch anything we missed and verify that events are properly being propagated through the system.

### Acceptance criteria
- [ ] The [Wiki documentation](https://github.com/newrelic/developer-website/wiki/Editing-Event-Data-and-Instrumentation) is updated to include any new events from I/O.
- [ ] Ensure that pack installs can be tied to new user signups, work with Zoe to confirm this.
- [ ] Fill in any missing events (NOTE: we might want to time box this ticket because of this AC)

jpvajda:
I've gone ahead and updated the WIKI for event types https://github.com/newrelic/developer-website/wiki/Editing-Event-Data-and-Instrumentation#event-data-1

aswanson-nr:
Previous event work https://github.com/newrelic/developer-website/issues/1538

jpvajda:
@jpvajda will work with @roadlittledawn on what existing events there are and then I can define if we need to consider instrumenting anything else.  We can also confirm the naming convention we are use for event names.

jpvajda:
Starting this work now..

jpvajda:
I've followed up with Zoe in Slack to see if we can confirm that quickstart installs can be tied to new user signups.

PR #1623, opened 2021-09-14
Update tessen naming for required format, consistency

roadlittledawn:
## Description

Tessen requires that event names be `camelCase` and category to be `TitleCase`. If they're not, events do not report / fail silently. Some of these events were not sending until I changed the format. I confirmed that all instrumented events sent when running locally.

I also updated the event names to use one standard name. That makes it easier to query downstream when Tessen packages up events for segment and NR1.

## Related issues / PRs
Partially addresses https://github.com/newrelic/developer-website/issues/1603

## Additional context
We don't quite seem to stick to a standard for tessen event naming. But generally, it seems `eventName` is the more broad of the two and `category` is a specific type of said event.

Our tessen wrapper uses this signature:

```
Tessen.track("eventName", "CategoryOfThing", {…})
```

Which packages that into this format in segment:

```
TDEV_CategoryOfThing_eventName
```

### DIFF
--- a/src/pages/instant-observability.js
+++ b/src/pages/instant-observability.js
@@ -92,7 +92,7 @@ const QuickstartsPage = ({ data, location }) => {
     setCategory(categoryParam);
 
     if (searchParam || filterParam || categoryParam) {
-      tessen.track('InstantObservability', 'QuickstartsCatalog', {
+      tessen.track('instantObservability', 'CatalogSearch', {
         filter: filterParam,
         search: searchParam,
         category: categoryParam,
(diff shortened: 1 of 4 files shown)


src/pages/instant-observability.js
```diff
@@ -92,7 +92,7 @@ const QuickstartsPage = ({ data, location }) => {
     setCategory(categoryParam);
 
     if (searchParam || filterParam || categoryParam) {
-      tessen.track('InstantObservability', 'QuickstartsCatalog', {
+      tessen.track('instantObservability', 'CatalogSearch', {
```
jpvajda, line 95:
```suggestion
      tessen.track('instantObservability', 'QuickstartCatalogSearch', {
```

jpvajda (changes requested):
had one suggestion to be consistent in naming

aswanson-nr (approved):

jpvajda (approved):

PR #1640, opened 2021-09-15
Add useEffect and state to conditionally render footer width/logo

LizBaker:
this is _a_ solution, not sure if it's _the_ solution. because it's an SSR issue i wasn't sure of a solution that doesn't involve the developer icon flashing in the footer (but you would have to be scrolled to the bottom and refresh to see it), same with the footer width/icon in the details view (scroll to the bottom and refresh to see)

### DIFF
--- a/src/@newrelic/gatsby-theme-newrelic/components/Logo.js
+++ b/src/@newrelic/gatsby-theme-newrelic/components/Logo.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/react';
 import IOLogo from '../../../components/IOLogo';
@@ -41,10 +41,15 @@ const Logo = ({ className, width }) => {
     <IOLogo className={className} width={width} />
   );
 
-  return window.location &&
-    window.location.href.includes('instant-observability')
-    ? instantObservabilityLogo
-    : developerLogo;
+  const [logo, setLogo] = useState(developerLogo);
+
+  useEffect(() => {
+    if (window.location.pathname.includes('instant-observability')) {
+      setLogo(instantObservabilityLogo);
+    }
+  }, []);
+
+  return logo;
 };
 
 Logo.propTypes = {
--- a/src/layouts/QuickStartLayout.js
+++ b/src/layouts/QuickStartLayout.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import {
   Layout,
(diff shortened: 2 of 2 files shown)


jpvajda:
I think this ok for GA, and we can look improving it post GA, or waiting until we move to a stand alone Gastby site for I/O which should resolve it I believe.

src/@newrelic/gatsby-theme-newrelic/components/Logo.js
```diff
@@ -41,10 +41,15 @@ const Logo = ({ className, width }) => {
...
   );
 
-  return window.location &&
-    window.location.href.includes('instant-observability')
-    ? instantObservabilityLogo
-    : developerLogo;
+  const [logo, setLogo] = useState(developerLogo);
+
+  useEffect(() => {
+    if (window.location.pathname.includes('instant-observability')) {
```
zstix, line 47: We _might_ also want to check if `window` exists so that this doesn't yell at us at build time.

src/layouts/QuickStartLayout.js
```diff
@@ -8,19 +8,15 @@ import PropTypes from 'prop-types';
...
-};
-
 const QuickStartLayout = ({ children }) => {
+  const [sidebarWidth, setSidebarWidth] = useState(0);
+
+  useEffect(() => {
+    if (window.location.pathname === '/instant-observability/') {
+      setSidebarWidth(300);
+    } else setSidebarWidth(0);
+  }, [children]);
```
zstix, line 18: Out of curiosity, do we need to add `children` to the dependency array? This might cause unnecessary calls to this function (whenever anything inside this component changes, this will get called).

zstix (commented):
Couple of comments.

Overall, while it's not ideal, this is a totally acceptable approach. If we decide to break this site out into it's own stand-alone site then this will not be an issue.

Thanks!!

src/layouts/QuickStartLayout.js
LizBaker: because this is in the layout it won't fire the effect again (it didn't seem) when switching to the details view since the layout doesn't change. using the children of the layout to determine if it fires was the only way i found to re-trigger it. very open to other solutions though! i just haven't found them 😭

src/@newrelic/gatsby-theme-newrelic/components/Logo.js
LizBaker: i believe window is always available in useEffect because it only runs after the render and not server side, but we could put a check in there for extra safety!

aswanson-nr (approved):

PR #1685, opened 2021-09-22
chore: aligns tags

jpvajda:
- moves the doc tag down under the description on the datsource tab
- removes an empty line from the overview file


rudouglas (approved):
👁️ :godmode: 🥈 👨‍🎨

PR #1688, opened 2021-09-22
Ru/utm parameters

rudouglas:


jpvajda:
We got the go-ahead from Andy on this approach for GA, so when the code is reviewed 🚢 🇮🇹

src/components/InstallButton.js
```diff
@@ -10,21 +10,61 @@ import {
...
+ * @param {Object} location
+ * @returns {String}
+ */
+const checkUtmParameters = (location) => {
+  const { state } = location;
+  if (!state) {
+    return false;
+  }
+  const { search: parameters } = state.prevPath;
+  const parametersArray = strippedParameters(parameters);
```
aswanson-nr, line 38: [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) will make the parsing and moving around search/query parameters much easier than dealing with the raw string.

gatsby-browser.js
```diff
@@ -54,5 +54,16 @@ const onClientEntry = () => {
   window.React = require('react');
   window.ReactDOM = require('react-dom');
 };
+const onRouteUpdate = ({ location, prevLocation }) => {
```
aswanson-nr, line 57: I was able to confirm with Andy that the UTM links will be directly to the quickstart itself, instead of to the I/O marketplace. Assuming this code was added to deal with that situation, I think we can remove it. If not, please ignore this comment!

src/components/PackTile.js
```diff
@@ -41,7 +42,9 @@ const PackTile = ({
         publicCatalogView: view,
         quickstartName: name,
       });
-      navigate(fields.slug);
+      navigate(fields.slug, {
+        state: { prevPath: location },
```
aswanson-nr, line 46: This is why I assumed the `onRouteUpdate` change was to pass the UTM parameters from the I/O marketplace to the details of a quickstart.

aswanson-nr (commented):
This looks great! I just had a few comments.

src/components/InstallButton.js
```diff
@@ -10,24 +10,56 @@ import {
   NR1_LOGIN_URL,
   NR1_GUIDED_INSTALL_NERDLET,
   NR1_PACK_DETAILS_NERDLET,
+  UTM_PARAMETERS,
+  NR1_SIGNUP_URL,
 } from '../data/constants';
 import { quickstart } from '../types';
 
+/**
+ * @param {Object} location
```
aswanson-nr, line 19:
```suggestion
 * @param {Object} parameters
```

src/components/InstallButton.js
```diff
@@ -10,24 +10,56 @@ import {
...
+ * @returns {String}
```
aswanson-nr, line 20:
```suggestion
 * @returns {Boolean}
```

src/components/InstallButton.js
```diff
@@ -10,24 +10,56 @@ import {
...
-    `?return_to=${encodeURIComponent(platformUrl)}`,
-    NR1_LOGIN_URL
-  );
 
-  return url.href;
+  const installUrl = new URL(hasUtmParameters ? NR1_SIGNUP_URL : NR1_LOGIN_URL);
+  parameters &&
+    parameters.forEach((value, key) => {
+      installUrl.searchParams.set(key, value);
+    });
```
aswanson-nr, line 59: While this is a really nice pattern in JSX to quickly show/hide something, it can be confusing when used within normal JS. If we use a quick if statement here it really clears things up.
```suggestion
  if (parameters) {
    parameters.forEach((value, key) => {
      installUrl.searchParams.set(key, value);
    });
  }
```

aswanson-nr (changes requested):
This looks fantastic! I just had a few very small nits

aswanson-nr (approved):

Issue #1733, opened 2021-09-30
[Public Catalog] I/O Tessen event changes


jpvajda:
Change event types to be consistent in naming convention. We currently have 2 events following the wrong pattern

### Acceptance Criteria 

- [x] ~~Change TDEV_InstantObservability_QuickstartViewToggle `to` TDEV_QuickstartViewToggle_InstantObservability~~
- [x] ~~Change TDEV_InstantObservability_QuickstartsCatalog `to` TDEV_QuickstartsCatalog_InstantObservability~~
- [x] Given we are only using 1 event type for the` install button` called `TDEV_QuickstartInstall_instantObservability` We want to  add a field to the event to pass in the name of the button that was clicked, either `install this quickstart` OR `see installation docs`
- [x] [Update WIKI ](https://github.com/newrelic/developer-website/wiki/Editing-Event-Data-and-Instrumentation#dev-site-event-data--instrumentation)with these changes

jpvajda:
we found that TDEV_QuickstartCatalogSearch_instantObservability is actually being used to track catalog searches, so I updated the wiki

we also didn't instruement the list view / grid view toggle as I recall we deemed as not worthy of tracking

PR #1735, opened 2021-09-30
chore: add quickstartbuttontext field to tessen

tabathadelane:
Moved tessen.track for the Install Button into the InstallButton.js. Added  a field to tessen to check the button text when clicked. 

Closes https://github.com/newrelic/developer-website/issues/1733


aswanson-nr (approved):
Saw some 404 weirdness in the preview, but a local run worked fine. I think the issue is on the hosting side. Looks good to me!

PR #1741, opened 2021-10-01
Replace beta with preview in callout

LizBaker:
Closes https://github.com/newrelic/newrelic-quickstarts/issues/414

### DIFF
--- a/src/components/quickstarts/BetaBanner.js
+++ b/src/components/quickstarts/BetaBanner.js
@@ -5,13 +5,13 @@ import { Callout } from '@newrelic/gatsby-theme-newrelic';
 const BetaBanner = () => (
   <Callout
     variant={Callout.VARIANT.COURSE}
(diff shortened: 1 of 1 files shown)


nr-kkenney (approved):
LGTM. thoughts on renaming to `PreviewBanner`? :trollface:

LizBaker:
@nr-kkenney I thought about it 😂 felt like we'll be dumping the whole component soon anyway 🤷‍♀️

jpvajda:
@LizBaker I think keeping it as a betabanner is more reflective of what it does, so the name is cool. Perhaps we can re-use this in the future for other beta callouts. so I'll merge this in now

If this component only on the devsite? or is it part of the global theme? I could see it being useful on the docs site eventually. cc @roadlittledawn

jpvajda (approved):

PR #1750, opened 2021-10-04
chore: instrumenting the banner button 

jpvajda:
This PR is adding instrumentation to different UI elements on I/O

1. Instrument the banner button click
2. TBD
3. TBD
4. TBD

### DIFF
--- a/src/components/IOBanner.js
+++ b/src/components/IOBanner.js
@@ -1,10 +1,26 @@
 import React from 'react';
 import { css } from '@emotion/react';
 import PropTypes from 'prop-types';
-import { Button, Link } from '@newrelic/gatsby-theme-newrelic';
+import {
+  Button,
+  Link,
+  useTessen,
+  useInstrumentedHandler,
+} from '@newrelic/gatsby-theme-newrelic';
 import ioBanner from '../images/ioBanner.png';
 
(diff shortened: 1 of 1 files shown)


src/components/IOBanner.js
```diff
@@ -1,10 +1,26 @@
...
+  const tessen = useTessen();
+
+  const handleBannerButtonClick = useInstrumentedHandler(
+    () => {
+      tessen.track('instantObservability', 'BannerButtonClick', {});
+    },
+    {
+      actionName: 'BannerButtonClick',
+    }
+  );
```
roadlittledawn, line 22: i don't think you need to use the `useInstrumentedHandler` hook here. i think you can just use tessen inside an inline function in the `onClick`

jpvajda: How do we decide when this used / not used?

jpvajda: After talking to @roadlittledawn I removed it.

jpvajda:
@rudouglas I'm curious if you can provide some direction how to setup tessen on the constants.js file for the support and explorer links, I put some commented out code in that file to call tessen but not sure how to actual set tessen.track on those links as they are inside a markdown block. 🤔

aswanson-nr (approved):

PR #1775, opened 2021-10-13
Update nr1-common.mdx

rudouglas:


alexronquillo:
Hey @rudouglas  Our workflow for these has transitioned to scraping the CLI to somewhat automate these docs. So this would get overwritten the next time we do that. When I just ran the scraper to see what the new docs would look like, I get:

````
Fetches data from new relic databases using a nrql (new relic query language) query

### Usage

```sh
nr1 nrql
```

### Options

|||
|---|---|
| `-a`, `--account=account` | (required) user account ID |
| `-q`, `--query=query` | (required) NRQL query to run |
| `-u`, `--ugly` | outputs the content without tabs or spaces |
| `--profile=profile` | the authentication profile you want to use |
| `--verbose` | adds extra information to the output |

### Description

See https://docs.newrelic.com/docs/insights/nrql-new-relic-query-language
````

I noticed the docs in the description results in a 404. I think the better place to update this link would be in the `nr1` CLI.

PR #1791, opened 2021-10-18
feat: hardcode installplan for codestream

rudouglas:
Do not merge until Codestream launch, 21st Oct 8:00 am PST


aswanson-nr (commented):
Let's remove the changes to `quickstarts.json` since this is now managed on the `main` branch of the repository. When we release this change, you can go in and manually trigger the workflow to pull in the codestream quickstart

PR #1794, opened 2021-10-18
feat: add quickstart name and id to tessen events

rudouglas:


src/components/Tabs/BarItem.js
```diff
@@ -7,7 +7,15 @@ import {
...
 import useTabs from './useTabs';
 
-const BarItem = ({ index, children, id, count, disabled }) => {
+const BarItem = ({
+  index,
+  children,
+  id,
+  count,
+  disabled,
+  quickstartName,
```
aswanson-nr, line 16: I feel a little icky about passing all this quickstart specific information into this more general component, even though this is the only page that currently uses it. If we avoid it now, it will be much easier to generalize this component later. That said, this is by no means a required change.
We could keep this component general by passing in a callback for click events, then that callback could be created with all the quickstart specific info in `QuickstartDetails` before even initializing the `BarItem` component.
```js
const tessenTrack = (action, quickstart) => (extraFields) => {
  tessen.track('instantObservability', action, {
    quickstartTitle: quickstart.title,
    quickstartId: quickstart.id,
    ...extraFields,
  };
};
```
Then you could call this with the relevant information and then pass the returned function to the downstream components.

src/components/Tabs/BarItem.js
```diff
@@ -1,37 +1,22 @@
...
 import PropTypes from 'prop-types';
 import { css } from '@emotion/react';
-import {
-  useTessen,
-  useInstrumentedData,
-} from '@newrelic/gatsby-theme-newrelic';
 import useTabs from './useTabs';
 
-const BarItem = ({ index, children, id, count, disabled }) => {
+const BarItem = ({ index, children, id, count, disabled, ...props }) => {
```
aswanson-nr, line 6: You can totally rename during deconstruction!
```suggestion
const BarItem = ({ index, children, id, count, disabled, onClick: onTabClick }) => {
```

src/components/quickstarts/SupportSection.js
```diff
@@ -1,21 +1,21 @@
...
+import { Link } from '@newrelic/gatsby-theme-newrelic';
 import {
   SUPPORT_LINK,
   COMMUNITY_LINK,
   QUICKSTART_SUPPORT_LEVELS,
 } from '../../data/constants';
 
-const SupportSection = ({ supportLevel }) => {
-  const tessen = useTessen();
+const SupportSection = ({ supportLevel, ...props }) => {
```
aswanson-nr, line 11: This one could be done in the same way

src/components/Tabs/BarItem.js
rudouglas: Yes totally agree on this one, i've refactored but not sure if that's as tidy as it could be 😬

PR #1815, opened 2021-10-21
Add multiple filter select

LizBaker:
WIP


LizBaker:
not sure why the linter is failing here. it doesn't fail locally and it seems to be mad about code that's been removed... that or it's complaining about the filter array method?

Issue #1800, opened 2021-10-19
[Public catalog] Move quickstart logo and CTAs, make header sticky


LizBaker:
We want the CTA buttons for quickstarts to remain visible to users even after they scroll. 

Mobile approach: keep the same approach we already have in place.

### AC
- [x] (logo + name + main CTA + breadcrumb) are in a sticky header and as the user scrolls down, they can access the main actions at any given time.
- [x] Logo moves to the left
- [x] CTA buttons move to the right 
- [x] As the screen size is reduced, the overall size of the sticky header section reduces
- [ ] Measure before and after for data analysis so we understand the impact of this change + / -
- [x] ~~the mobile view should also have this treatment~~  -`won't do`

### Design
![SKX929djFq0XEe7b0dVskeQPsHqVb0YEe4lwOHLZJAGOzTa63g-fIPBQrCoANf4vBYLyL4kDOs-Si2sSlsTeYLaoKA3UvP5pumOTojAGrakilADT3kNvFVl3KBGZaw61oHN-vyFZ=s0](https://user-images.githubusercontent.com/39655074/138000123-f33a9045-ffe0-4d73-8024-322e746b7cf2.png)

jpvajda:
We are waiting on some final input from @legnaleama to finish this up

PR #1833, opened 2021-10-28
make quickstart header sticky

nr-kkenney:
## Summary

* reorganized the header to have logo on the left, title & summary in
the middle, and CTA buttons on the right
* header is sticky and will remain on page when a user scrolls. it
will 'stack' up  with the global header which is also sticky

## Screenshots
Screen Shot 2021-10-29 at 2 26 06 PM
Screen Shot 2021-10-29 at 2 26 15 PM
Screen Shot 2021-10-29 at 2 26 27 PM

## Links
Resolves: #1800

### DIFF
--- a/src/templates/QuickstartDetails.js
+++ b/src/templates/QuickstartDetails.js
@@ -128,19 +128,18 @@ const QuickstartDetails = ({ data, location }) => {
 
               @media (min-width: 760px) {
                 background: var(--primary-background-color);
-                border: 1px solid var(--border-color);
+                border-bottom: 1px solid var(--border-color);
                 border-radius: 0.25rem;
-                box-shadow: 0 2px 4px rgba(0, 0, 0, 0.1);
                 grid-template-areas:
                   'logo title cta'
                   'logo summ cta';
-                padding: 8px 0;
+                padding: 16px 0 24px;
                 position: sticky;
                 top: var(--global-header-height);
                 z-index: 80;
               }
 
-              .dark-mode & {
+              .dark-mode {
                 box-shadow: none;
               }
             `}
--- a/src/templates/QuickstartDetails.js
+++ b/src/templates/QuickstartDetails.js
@@ -107,10 +107,8 @@ const QuickstartDetails = ({ data, location }) => {
               )
             }
             css={css`
-              background: var(--primary-background-color);
               border-bottom: none;
               display: grid;
-              padding-bottom: 0;
               grid-template-areas:
                 'logo title cta'
                 'logo summ cta';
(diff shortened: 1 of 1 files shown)


nr-kkenney:
@jpvajda if you have a moment to take a look. ive addressed the comments, so should be good for your eyes 🙈

jpvajda:
I foresee several little nits in this review, so I think it may be most optimal to hop on a quick zoom session with @legnaleama next week to work through those. 

I have some feedback /questions but in this case I think a quick working session would be easier

src/templates/QuickstartDetails.js
```diff
@@ -109,17 +109,40 @@ const QuickstartDetails = ({ data, location }) => {
...
+                grid-template-areas:
+                  'logo title cta'
+                  'logo summ cta';
+                padding: 8px 0;
+                position: sticky;
+                top: var(--global-header-height);
+                z-index: 80;
+              }
+
+              .dark-mode & {
```
clarkmcadoo, line 143: I think you can do without this `&`, we're just trying to target the `dark-mode` class here 👍

jpvajda (dismissed):
I'd like to schedule some time with Angel, to discuss any small corrections to this, I see a few but it will be easier to talk through them on quick chat

legnaleama:
Here are some minor adjustments:
- Replace current border and box shadow for `border-bottom: 1px solid var(--border-color);` (We'll improve this on a future PR)
- Adjust header's padding to `padding: 16px 0 24px;`

jpvajda:
@nr-kkenney feedback is in. 🎨

nr-kkenney:
@clarkmcadoo @jpvajda @legnaleama changes made 👍

src/templates/QuickstartDetails.js
```diff
@@ -109,17 +109,39 @@ const QuickstartDetails = ({ data, location }) => {
...
               grid-column-gap: 1rem;
               grid-row-gap: 1rem;
+              grid-template-areas:
+                'title logo'
+                'summ logo'
+                'cta logo';
+              justify-content: normal;
+              justify-self: center;
               row-gap: 1rem;
+              width: 101%;
```
zstix, line 121: Setting the width to _more_ than `100%` is kinda weird, but I totally see why you're doing it. I tried playing around with the page layout to try and get the underlying content to be properly hidden underneath the header and couldn't come up with anything.

It would be nice to circle back to this at some point, but I don't think we should hold up this PR.

zstix (approved):

clarkmcadoo (approved):

src/templates/QuickstartDetails.js
nr-kkenney: yea if im remembering this correctly, it was just a pixel of the left most point of the W character that wasn't being hidden 🙈

Issue #1692, opened 2021-09-22
[Public Catalog] Add a Related resources  section to the quickstart details


jpvajda:
`Problem:` User’s don’t know what else they can do with New Relic when viewing a quickstart externally.

### Acceptance Criteria 

- [x] Show related content on a quickstart details page.
- [x] Use existing related resources component [available in the theme](https://github.com/newrelic/gatsby-theme-newrelic/tree/develop/packages/gatsby-theme-newrelic#relatedresources)
- [x] the placement of the component would be within `the how to use this quickstart` component
- [x] align the width of  `the how to use this quickstart` similar to how it looks on the Doc Site or DevSite today.
- [x] return a total of 5 results in the related resources.

### screenshot

<img width="482" alt="Screen Shot 2021-10-27 at 11 17 03 AM" src="https://user-images.githubusercontent.com/1395158/139114686-a9847ca4-70c1-46a7-b918-6ab0d4095dec.png">

PR #1841, opened 2021-11-02
Add related resources to quickstart details

LizBaker:
Closes https://github.com/newrelic/developer-website/issues/1692

Screen Shot 2021-11-04 at 4 52 41 PM


gatsby-config.js
```diff
@@ -59,10 +60,23 @@ module.exports = {
             refetch: Boolean(process.env.BUILD_RELATED_CONTENT),
             engineKey: 'Ad9HfGjDw4GRkcmJjUut',
             limit: 5,
-            getSlug: ({ node }) => node.frontmatter.path,
+            getSlug: ({ node }) => {
+              if (node.internal.type === 'Mdx') {
+                return node.frontmatter.path;
+              } else if (node.internal.type === 'Quickstarts') {
```
clarkmcadoo, line 66: was an `else if` used instead of just an `else` because we might be adding more `node.internal.type`s eventually?

clarkmcadoo (dismissed):

gatsby-config.js
LizBaker: mostly just to be extra explicit for the heck of it 😂

jpvajda (approved):
This looks great now! If all code checks out, let's ship it!

PR #1864, opened 2021-11-05
Update tile design

jerelmiller:
## Description

The guide tiles on the pages needed a bit of a refresh. On the [docs site](), we use a solid tile to launch into other parts of the docs. This site felt like it needed the same treatment.

This PR aligns the guide tile design a bit more with the docs tiles. Specifically it update to use the same solid background treatment, updates the title size to match the docs site, and moves the duration to the bottom of the tile. This provides more room for the title before it wraps.

This PR also addresses the inconsistencies with the duration given for a guide. There were some places that used `5 min`, and others that spelled out the minutes (i.e. `5 minutes`), which looked kinda bad. Now the frontmatter accepts an integer instead and inserts the label itself.

## Screenshot(s)

**Before**

Screen Shot 2021-11-04 at 5 33 09 PM
Screen Shot 2021-11-04 at 5 33 16 PM

**After**
Screen Shot 2021-11-04 at 5 36 17 PM
Screen Shot 2021-11-04 at 5 36 24 PM


nr-kkenney:
looks good to me. @jpvajda this look good to you as well?

nr-kkenney (approved):

jpvajda:
This is awesome! Thanks @jerelmiller. 🥳

PR #1866, opened 2021-11-05
Add background for quickstarts in dark mode

zstix:
## Description

These were looking a little rough in dark-mode (see before / after screenshots in the pull request). This isn't a perfect solution, but I think it looks much better than before.

## Screenshot(s)

**BEFORE**
Screen Shot 2021-11-05 at 10 45 47

**AFTER**
Screen Shot 2021-11-05 at 10 52 06
*

### DIFF
--- a/src/components/PackTile.js
+++ b/src/components/PackTile.js
@@ -86,8 +86,7 @@ const PackTile = ({
           margin: 0 auto 10px;
 
           .dark-mode & {
-            background-color: rgb(231 231 231 / 0);
-            background: white;
+            background-color: white;
           }
 
           ${view === VIEWS.LIST &&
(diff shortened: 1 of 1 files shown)


legnaleama:
While adding pure white can cause some visual strain on dark mode because of the high contrast, I think we can leave it as it is for now given the low traffic. 

That said, I'm reviewing our overall visual treatment for Dark mode and I'm working on documenting a set of improvements for Q4. No action on that for now though.

clarkmcadoo (approved):

jpvajda:
@legnaleama Thanks!  this improvement actually make the overall exp. a little better, so I'm going to merge this in, but I'm looking forward to whatever you come up with.

PR #1871, opened 2021-11-08
feat: add notification category

rudouglas:
Adding a notification as per Slack request

### Screenshot 
Screen Shot 2021-11-17 at 12 09 27 PM

### DIFF
--- a/src/data/instant-observability-categories.json
+++ b/src/data/instant-observability-categories.json
@@ -96,7 +96,6 @@
     "displayName": "Notifications",
     "associatedKeywords": [
       "notifications",
-      "outbound integrations",
     ]
   }
 ]
--- a/src/data/instant-observability-categories.json
+++ b/src/data/instant-observability-categories.json
@@ -97,7 +97,6 @@
     "associatedKeywords": [
       "notifications",
       "outbound integrations",
-      "destinations"
     ]
   }
 ]
--- a/src/data/instant-observability-categories.json
+++ b/src/data/instant-observability-categories.json
@@ -90,5 +90,14 @@
     "associatedKeywords": [
       "mlops"
     ]
+  },
+  {
+    "value": "notifications",
+    "displayName": "Notifications",
+    "associatedKeywords": [
+      "notifications",
+      "outbound integrations",
+      "destinations"
+    ]
   }
 ]
--- a/src/data/instant-observability-categories.json
+++ b/src/data/instant-observability-categories.json
@@ -7,95 +7,66 @@
   {
     "value": "featured",
     "displayName": "Featured",
-    "associatedKeywords": [
-      "featured"
-    ]
+    "associatedKeywords": ["featured"]
   },
   {
     "value": "application-monitoring",
     "displayName": "Application monitoring",
-    "associatedKeywords": [
-      "language agent"
-    ]
+    "associatedKeywords": ["language agent"]
   },
   {
     "value": "infrastructure-and-os",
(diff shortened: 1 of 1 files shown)


jpvajda:
closing as this should come from the internal I/O categories, and not just be set on the external site only

jpvajda:
It sounds like DevEx is implementing this new catagory so re-opening

zstix (dismissed):

src/data/instant-observability-categories.json
```diff
@@ -90,5 +90,14 @@
...
     ]
+  },
+  {
+    "value": "notifications",
+    "displayName": "Notifications",
+    "associatedKeywords": [
+      "notifications",
+      "outbound integrations",
+      "destinations"
+    ]
```
jerelmiller, line 101: Per the thread, shouldn't this just be `notifications`?

jpvajda: @jerelmiller We were looking for confirmation from what is going to be defined internally, so if that is the case, do you mean the only associated keyword is `notifications` and the others should be removed?

jerelmiller: @jpvajda correct according to [this message](https://newrelic.slack.com/archives/C01R2TB8BUZ/p1637160342242100?thread_ts=1636378774.153100&cid=C01R2TB8BUZ)

jpvajda: 👍

jerelmiller:
We've got a PR opened for in-product as well, so we can get that merged whenever you all are ready here.

src/data/instant-observability-categories.json
```diff
@@ -90,5 +90,14 @@
     "associatedKeywords": [
       "mlops"
     ]
+  },
+  {
+    "value": "notifications",
+    "displayName": "Notifications",
+    "associatedKeywords": [
+      "notifications",
+      "outbound integrations",
```
jpvajda, line 99:
```suggestion

```

src/data/instant-observability-categories.json
```diff
@@ -90,5 +90,14 @@
...
+      "destinations"
```
jpvajda, line 100:
```suggestion

```

src/data/instant-observability-categories.json
jpvajda: confirmed this isn't required.

src/data/instant-observability-categories.json
jpvajda: confirmed this isn't required.

jpvajda (changes requested):
changes suggested to align with the associated keywords.

jpvajda:
This is failing gatsby build so I'm taking a look at it.

src/data/instant-observability-categories.json
```diff
@@ -90,5 +90,12 @@
     "associatedKeywords": [
       "mlops"
     ]
+  },
+  {
+    "value": "notifications",
+    "displayName": "Notifications",
+    "associatedKeywords": [
+      "notifications",
```
jerelmiller, line 98:
```suggestion
      "notifications"
```

jerelmiller: I think the trailing comma is what is causing it to fail since JSON doesn't support trailing commas

jpvajda: That's the culprit.. 🥇

jerelmiller (approved):
Thanks for getting this in!

PR #1875, opened 2021-11-08
chore: added script to dev site home

jpvajda:
This PR attempts to add the crazy egg script to the developer site home page as well, so both the `/` page an `/instant-observability/` both have the script running.

### DIFF
--- a/src/components/DevSiteSeo.js
+++ b/src/components/DevSiteSeo.js
@@ -20,7 +20,7 @@ function DevSiteSeo({ description, meta, title, tags, location, type }) {
 
   const crazyEgg = (location) => {
     if (
-      location.pathname === '/index.html' ||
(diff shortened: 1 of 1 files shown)


jpvajda:
This approach also seems to throw the same errors as using a ternary operator...

Uncaught Error: CE: multiple userscripts installed
    <anonymous> http://script.crazyegg.com/pages/scripts/0045/9836.js:1
9836.js:1:70
    <anonymous> http://script.crazyegg.com/pages/scripts/0045/9836.js:1

jpvajda:
I've added a path for index.html, but it seems the script is still not firing on the index page in local testing

LizBaker (approved):

aswanson-nr (approved):

PR #1893, opened 2021-11-17
fix: tessen event not triggered on click for guided install tile

abrunner94:
## Description

Looks like `navigate()` prevents Tessen from registering that event if it happens at the same time as `tessen.track`. Using `useInstrumentHandler` is a better approach compared to using timeouts.

### DIFF
--- a/src/experiments/super_tiles/components/GuidedInstallTile.js
+++ b/src/experiments/super_tiles/components/GuidedInstallTile.js
@@ -10,7 +10,7 @@ import { SIGNUP_LINK } from '../../../data/constants';
 
 const GuidedInstallTile = () => {
   const handleButtonClick = useInstrumentedHandler(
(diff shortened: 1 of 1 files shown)


zstix (changes requested):
I think we can use our custom hook to handle the instrumentation for the button click: [`useInstrumentedHandler`](https://github.com/newrelic/gatsby-theme-newrelic/tree/main/packages/gatsby-theme-newrelic#useinstrumentedhandler).

abrunner94:
> I think we can use our custom hook to handle the instrumentation for the button click: [`useInstrumentedHandler`](https://github.com/newrelic/gatsby-theme-newrelic/tree/main/packages/gatsby-theme-newrelic#useinstrumentedhandler).

Nice - thanks for this one!

zstix (approved):

PR #1904, opened 2021-11-19
feat: Add OpenTelemetry lab

alexronquillo:
## Description

Adds a lab for OpenTelemetry based on a Data Nerd Days 2.0 talk.


aswanson-nr:
We recently merged https://github.com/newrelic/developer-website/pull/1916, which changed the routing from the frontmatter `path` field to use the file path after `src/markdown-pages`. The transition should be as easy as making sure the file is in the right spot on the filesystem and deleting the `path` field. Please update your branch with the latest from the `develop` branch or you will probably see some build errors! Thank you.

src/markdown-pages/collect-data/opentelemetry/index.mdx
```diff
@@ -0,0 +1,39 @@
+---
+title: 'Instrument and Analyze OpenTelemetry data'
+template: 'LabOverviewTemplate'
+description: 'Instrument and Analyze OpenTelemetry data'
+---
+
+<Intro>
+
+You're a developer who recently released a simple .NET weather forecast application to the world, and people can't stop talking about it! Surprised by the spike in popularity, you decided you should start collecting metrics and traces from your application so you can analyze them in New Relic.
```
mehreentahir16, line 9: Is it necessary to mention "simple"? 
It reads to me like this guide will only work with simple applications and not the complicated ones.

src/markdown-pages/collect-data/opentelemetry/instrument.mdx
```diff
@@ -0,0 +1,404 @@
...
+
+```bash
+export OTEL_EXPORTER_OTLP_ENDPOINT=<REGION-BASED-OTLP-ENDPOINT>
+export NR_LICENSE_KEY=<YOUR-NEW-RELIC-LICENSE-KEY>
+```
+
+The [OpenTelemetry Protocol](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md) (OTLP) endpoint is the url of our OpenTelemetry receiver. Your service sends data directly to New Relic through this endpoint. [We have a US endpoint and an EU endpoint](https://docs.newrelic.com/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/opentelemetry-quick-start/#review-settings), so make sure you choose the appropriate one based on your region.
+
+There are several different types of API keys to choose from in New Relic that each serve a different purpose. To instrument your application with OpenTelemetry, you need a [license key](https://docs.newrelic.com/docs/apis/intro-apis/new-relic-api-keys/#ingest-license-key).
+
```
mehreentahir16, line 94: I think we should add a callout here to remind the reader to replace the end point and licence key in the above to make sure they export the right variables. Reminder can be useful

src/markdown-pages/collect-data/opentelemetry/set-up-env.mdx
```diff
@@ -0,0 +1,129 @@
...
+procIdx: 1
+---
+
+<Callout variant="course" title="lab">
+
+This procedure is part of a lab that teaches you how to instrument your application with OpenTelemetry. If you haven't already, check out the [lab introduction](/collect-data/opentelemetry).
+
+</Callout>
+
+Before you can walk through the lab proper, you need to set up your development environment. Here, you:
```
mehreentahir16, line 15:
```suggestion
Before you can walk through the lab properly, you need to set up your development environment. Here, you:
```

alexronquillo: the "lab proper" means the lab, specifically, or without extra context. Setting up part of the environment is a preliminary step and isn't considered part of the "lab proper" here. 

"properly" doesn't really make sense as an adverb in this context as we're not suggesting the walk through the lab properly vs improperly.

I'll think if there's another way to say this though.

jaesius (approved):

PR #1916, opened 2021-11-28
Update routing to use filepath instead of frontmatter.path to generate routes

jerelmiller:
## Description

We have too many MDX guides that are put in seemingly random locations on the file system. This makes it difficult to find and edit a file via the URL since you first have to either click the "Edit page" button, or use your editor's global find feature to find a path that matches the URL.

This PR seeks to eliminate the confusion and make locating a guide at a particular URL much more straight forward. Instead of letting the developer define a custom `path` on frontmatter, the filepath of the MDX file itself will determine the URL for the guide. This eliminates the `path` property altogether.

To ease the transition, a build error will occur with a helpful error in the console if a `path` property is detected alerting the developer that the property has been removed and a recommendation for where to move the filepath if it differs from the `path`.

All guides have been updated to remove the `path` property and all of them are located in the proper location.

## Reviewer Notes

**NOTE:** This WILL affect any open PRs that introduce new guides or update the `path`. If this gets merged beforehand, we will want to get these changes pulled into those branches and apply the updates, otherwise the build will fail.

Specifically, this looks like it affects these PRs (at the time of this writing):
* https://github.com/newrelic/developer-website/pull/1904


jerelmiller:
cc @alexronquillo @jaesius so you know about these changes 🙂

jpvajda:
I've added this the PR to the deven github project board for review. Thanks!

alexronquillo:
This is fine with me. I think we've never really defined how we were supposed to organize the filesystem since the paths were arbitrary, so over time we've sometimes placed them to match the path, sometimes we've organized them by topic (especially when we didn't know where they would eventually end up), etc. 

Thanks for all the work on this @jerelmiller having only one way to organize docs will certainly make decision-making easier on us :)

alexronquillo (approved):

aswanson-nr (approved):

PR #1941, opened 2021-12-09
Attempting to use `dotenv` to access non-GATSBY_ env variables

zstix:
## Description

I wanted to investigate if using `dotenv` (as [the docs](https://www.gatsbyjs.com/docs/how-to/local-development/environment-variables/) suggests) solves our issue accessing environment variables that aren't prepended with `GATSBY_` (and therefore exposed in the browser).

**Do not merge**.


zstix:
I give up.

PR #1945, opened 2021-12-13
[SSR] Add sort to query

LizBaker:
This adds a sort component which sets a sort value in the query params and navigates to that page. The main IO page then reloads and queries with the sort value from the params
Screen Shot 2021-12-13 at 3 57 57 PM


LizBaker:
closing as dupe for https://github.com/newrelic/developer-website/pull/1948

PR #1948, opened 2021-12-14
Campfire/sort fixup

aswanson-nr:
## Description
Quick PR to clean up some git stuff going on in #1945

This adds a sort component which sets a sort value in the query params and navigates to that page. The main IO page then reloads and queries with the sort value from the params

Screen Shot 2021-12-13 at 3 57 57 PM

### DIFF
--- a/src/components/quickstarts/QuickstartSort.js
+++ b/src/components/quickstarts/QuickstartSort.js
@@ -2,7 +2,7 @@ import React from 'react';
 import { navigate } from 'gatsby';
 import PropTypes from 'prop-types';
 
-const QuickstartSort = ({ className }) => {
+const QuickstartSort = ({ className, location }) => {
   const params = new URLSearchParams(location.search);
   const sortParam = params.get('sort');
   const handleChange = (event) => {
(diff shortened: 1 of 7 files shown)


src/components/quickstarts/QuickstartSort.js
```diff
@@ -0,0 +1,32 @@
...
+
+const QuickstartSort = ({ className }) => {
+  const params = new URLSearchParams(location.search);
+  const sortParam = params.get('sort');
+  const handleChange = (event) => {
+    params.set('sort', event.target.value);
+    navigate(`/instant-observability?${params.toString()}`);
+  };
+  return (
+    <>
```
aswanson-nr, line 13: Looks like we don't need this fragment!

aswanson-nr (commented):
Noticed an issue where `location` is available during render if you go directly to `/instant-observability` on a prod build. 
Screen Shot 2021-12-14 at 2 44 00 PM

Also it would be worth replicating the mobile tab dropdown styling for the sort select.

src/components/quickstarts/QuickstartSort.js
zstix: Looks like this comment got prematurely resolved.

zstix (approved):
Looks good. There are some areas that we might want to consider refactoring, but that doesn't need to happen in this PR.

PR #1949, opened 2021-12-15
Add search and filter via ssr

tabathadelane:
Adds search, category, and filter requests to the ssr page

Closes newrelic/instant-observability-website#42


LizBaker:
if anyone wants to continue between now and 1/3/22-- current status: last AC is adding tests

PR #1981, opened 2021-12-28
I/O Quickstart Search & Grid Controls

josephgregoryii:
New styling changes and refactoring for IO


rudouglas:
Had a comment on the UX here, which I don't think is caused directly by this PR, but when you click on a category, or search something in the I/O searchbox, the view jumps back to the top of the page which is very jarring. To recreate:

 - Go to the preview https://developerwebsitefeatureioredes-joeiosearchcategory.gtsb.io/
 - Scroll down the page a little
 - Click on a Category on the left
 - OR search for something in the I/O search box

josephgregoryii:
@rudouglas great catch - I'll try to work and see if i can get that change in this PR. If not i'll create a ticket for it 👍🏻

jpvajda:
This looks good to me, apart from what @rudouglas pointed out. 💯

jpvajda (approved):
the scrolling issue is now fixed.

gatsby-browser.js
```diff
@@ -55,4 +55,19 @@ const onClientEntry = () => {
   window.ReactDOM = require('react-dom');
 };
 
-export { wrapPageElement, onInitialClientRender, onClientEntry };
+const shouldUpdateScroll = ({ routerProps: { location } }) => {
+  const transitionDelay = 600;
```
clarkmcadoo, line 59: Can you put this variable in all caps since its an artificially set constant? like `TRANSITION_DELAY`;

josephgregoryii: fixed 🤠

src/components/SegmentedControl.js
```diff
@@ -1,79 +1,61 @@
...
-              if (item.disabled) return;
               setSelected(value);
               onChange && onChange(e, value);
             }}
+            css={css`
+              margin-left: 8px;
+            `}
           >
-            {get(item, 'label')}
+            <Icon
```
aswanson-nr, line 52: We should add a `title` to this icon so users know what it does when they mouse over it.

src/pages/instant-observability.js
```diff
@@ -413,47 +351,97 @@ const QuickstartsPage = ({ data, location }) => {
...
+                --svg-color: var(--color-neutrals-700);
+
+                svg {
+                  width: 16px;
+                  height: 16px;
+                  color: var(--svg-color);
+                }
+
+                .dark-mode & {
+                  --svg-color: var(--primary-text-color);
```
aswanson-nr, line 417: We most likely want to add some different background color in darkmode. The current color needs more contrast with the placeholder text:
Screen Shot 2022-01-03 at 2 12 52 PM

src/pages/instant-observability.js
```diff
@@ -660,38 +647,25 @@ const QuickstartsPage = ({ data, location }) => {
...
+            <span
               css={css`
-                min-width: 155px;
-                margin-left: 20px;
-                display: inline;
-
-                @media screen and (max-width: 1180px) {
-                  margin-left: 0px;
-                }
+                color: var(--primary-text-color);
```
aswanson-nr, line 663: We need to add a text color for darkmode, the current color blends in with the background:
Screen Shot 2022-01-03 at 2 15 34 PM

gatsby-browser.js
```diff
@@ -55,4 +55,19 @@ const onClientEntry = () => {
   window.ReactDOM = require('react-dom');
 };
 
-export { wrapPageElement, onInitialClientRender, onClientEntry };
+const shouldUpdateScroll = ({ routerProps: { location } }) => {
```
aswanson-nr, line 58: I'm a little confused as to what this is for... It looks like it resets the scroll position on any page without a `search` query parameter. Since it affects every page on the site, we might want to look at other ways of achieving what we're looking for.

clarkmcadoo (approved):

src/pages/instant-observability.js
```diff
@@ -814,9 +815,13 @@ const Label = ({ children, htmlFor }) => (
...
     css={css`
       display: block;
-      font-size: 12px;
-      font-weight: bold;
-      margin-bottom: 0.25rem;
+      font-size: 14px;
+      font-weight: 600;
+      margin-bottom: 1.5rem;
+
+      .light-mode & {
```
aswanson-nr, line 822: I don't think we denote light mode as a class, it's just the default. We should reverse the logic here and default to light mode colors.

src/pages/instant-observability.js
```diff
@@ -747,6 +735,19 @@ const QuickstartsPage = ({ data, location }) => {
...
             {filteredQuickstarts.map((pack) => (
               <PackTile
+                css={css`
+                  @media screen and (max-width: ${COLLAPSE_BREAKPOINT}) {
+                    flex-direction: row;
+
+                    img {
+                      display: none;
+                    }
+                    .tag-div {
```
aswanson-nr, line 745: This seems like something that could be broken out into the `PackTile` component itself, instead of passing it down via css classes. The breakpoints should work there as well

src/pages/instant-observability.js
```diff
@@ -722,6 +696,14 @@ const QuickstartsPage = ({ data, location }) => {
                 css={css`
                   ${view === VIEWS.GRID && `height: 100%;`}
                   background-color: var(--tertiary-background-color);
+
+                  @media screen and (max-width: ${COLLAPSE_BREAKPOINT}) {
+                    flex-direction: row;
+
+                    img {
```
aswanson-nr, line 703: This works for most tiles, but some tiles that do not have an image use a `div` instead. I would recommend moving this breakpoint logic to the `PackTile` component itself, that way it's all encapsulated and you can deal with the different types of display "images".

aswanson-nr (changes requested):
Looks great! Just noticed some things around breakpoints and dark mode!

gatsby-browser.js
josephgregoryii: I'd love to chat about this! There was some hesitation with this approach for that exact reason.

src/pages/instant-observability.js
josephgregoryii: update: 
Screen Shot 2022-01-03 at 3 18 16 PM

Does this work better?

aswanson-nr: Looks good to me!

src/pages/instant-observability.js
josephgregoryii: That was a weird testing thing i was doing (not sure why, now that i think of it 🤣). Consider it gone 👍🏻

josephgregoryii:
@aswanson-nr changes made!

src/pages/instant-observability.js
```diff
@@ -23,6 +23,8 @@ import CATEGORIES from '../data/instant-observability-categories';
 
 import SuperTiles from '../components/SuperTiles';
 
+const COLLAPSE_BREAKPOINT = '760px';
```
aswanson-nr, line 26: Since we're using this in multiple components now, we should break it out into the constants file. Maybe rename it to `QUICKSTARTS_COLLAPSE_BREAKPOINT`

aswanson-nr (commented):
One last comment!

src/pages/instant-observability.js
josephgregoryii: Oooo i like it. I'll get to it.

src/data/constants.js
```diff
@@ -50,3 +50,5 @@ export const RESERVED_QUICKSTART_IDS = {
   GUIDED_INSTALL: 'GUIDED_INSTALL',
   BUILD_YOUR_OWN_QUICKSTART: 'BUILD_YOUR_OWN_QUICKSTART',
 };
+
+export const COLLAPSE_BREAKPOINT = '760px';
```
aswanson-nr, line 54: Okay, I'm so sorry, but let's call this something with quickstarts in the name since that's the only place it's used

aswanson-nr (approved):

rudouglas:
awwwwwesome

PR #1989, opened 2022-01-05
remove auto-scroll for I/O

josephgregoryii:
changes are only applied to I/O

### DIFF
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -56,7 +56,7 @@ const onClientEntry = () => {
 };
 
 const shouldUpdateScroll = ({ routerProps: { location } }) => {
-  const PAGE_OFFSET_HEIGHT = 575;
+  const PAGE_OFFSET_HEIGHT = 310;
   const IS_IO = location.pathname.includes('instant-observability');
 
   // scrolls only after search query parameter on I/O site changes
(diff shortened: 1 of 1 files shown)


jpvajda (approved):

jpvajda:
from a product perspective this LGTM, if the code checks out in a review, :shipit:

gatsby-browser.js
```diff
@@ -55,4 +55,27 @@ const onClientEntry = () => {
   window.ReactDOM = require('react-dom');
 };
 
-export { wrapPageElement, onInitialClientRender, onClientEntry };
+const shouldUpdateScroll = ({ routerProps: { location } }) => {
+  const PAGE_OFFSET_HEIGHT = 575;
```
aswanson-nr, line 59: I think we may want to target the bottom of the banner for this, the current height slightly cuts off the top of the search bar.

gatsby-browser.js
```diff
@@ -55,4 +55,27 @@ const onClientEntry = () => {
   window.ReactDOM = require('react-dom');
 };
 
-export { wrapPageElement, onInitialClientRender, onClientEntry };
+const shouldUpdateScroll = ({ routerProps: { location } }) => {
+  const PAGE_OFFSET_HEIGHT = 310;
```
aswanson-nr, line 59: My final comment! Let's just add a comment here to explain this magic number

PR #1992, opened 2022-01-07
Clark/mobile io updates - 1964

clarkmcadoo:
Final design changes for I/O mobile

### DIFF
--- a/src/components/IOBanner.js
+++ b/src/components/IOBanner.js
@@ -47,7 +47,7 @@ const BannerHeaderContent = () => (
     <body
       css={css`
         background: none;
-        color: var(--color-brand-300);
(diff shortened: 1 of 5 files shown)


aswanson-nr:
Got some interesting behavior on at the image breakpoint
Screen Shot 2022-01-07 at 1 17 56 PM
Since we're no longer hiding the image, I think we can remove the whole block starting on line 95 of the `PackTile` component

aswanson-nr:
Seeing this with the iPad Air setting
Screen Shot 2022-01-07 at 1 29 48 PM

nr-kkenney (approved):
checked the different resolution sizes for devices, and i dont see what we saw before so looks good 👍

PR #2017, opened 2022-01-13
fix(dev site): remove insert key

zuluecho9:
## Description

From docs team jira: removing references to insert key and the 'x-insert-key' header and replacing with license key and 'Api-Key' header.


jpvajda (approved):

jpvajda:
@zuluecho9  I also wanted confirm if you still wanted this merged, as it's an older PR

PR #2019, opened 2022-01-14
feat: Created new 'data integration' page in contribute to quickstarts section

jcountsNR:
## Description

Adding a new page under the 'Contribute to Quickstarts' guide which should help partners be able to integrate their data easier.

## Screenshot(s)

Screen Shot 2022-01-13 at 4 15 35 PM

## Use Conventional Commits

Please help the maintainers by leveraging the following [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/)
standards in your pull request title and commit messages.

## Use `chore`

* for minor changes / additions / corrections to content.
* for minor changes / additions / corrections to images.
* for minor non-functional changes / additions to github actions, github templates, package or config updates, etc

```bash
git commit -m "chore: adjusting config and content"
```

## Use `fix`

* for minor functional corrections to code.

```bash
git commit -m "fix: typo and prop error in the code of conduct"
```

## Use `feat`

* for major functional changes or additions to code.

```bash
git commit -m "feat(media): creating a video landing page"
```


jcountsNR:
@jaesius , please review when you get time and let me know. We can set up time to review as well if needed.

alexronquillo:
Hey @jcountsNR! Thanks for submitting this piece! Before I edit and review it, I want to talk through its positioning on the dev site.

Here, we have essentially two types of content:

- **Guides:** standalone documents, tutorials, etc. that aim to teach users how to do a specific thing or explain how something works
- **Labs:** a collection of procedures that presents a guided walkthrough of a user story

What you've submitted wants to be both a guide and a lab procedure, and I don't think that's the best approach.

If I understand your intent correctly, you want to show users how to send events from their application because having data in New Relic is a prerequisite to building a quickstart. This is great. However, the lab you're trying to connect to ("Build a quickstart") starts with resources that already send data to New Relic, specifically:

> A python program [FlashDB] that mimics the database and also provides functions for create, read, update, and delete (CRUD) operations. **It also makes use of newrelic_telemetry_sdk to send telemetry data to New Relic.**

In other words, your piece would be an alternative starting point, except it doesn't work with data that's relevant to the narrative. Lab procedures need to be consistent with the narrative.

So, I recommend you either:

- **Option 1:** Create a new, optional lab procedure that comes before [Deploy your application](https://developer.newrelic.com/contribute-to-quickstarts/build-a-quickstart/deploy-your-application/). Call it something like "Instrument your application". In it, have the user start with a version of FlashDB that doesn't send data to New Relic. Next, have them add the instrumentation so they end up with the version of FlashDB that they use in "Deploy your application". This way, the rest of the narrative follows naturally. (This option would mean dropping the content you wrote and starting from scratch.)
- **Option 2:** Move the guide you wrote under the top-level "Collect data" nav item on the dev site. Drop all mentions of quickstarts (sending events isn't specific to quickstarts, anyway). Then, reference this guide from the quickstarts lab for users that aren't already sending data to New Relic. (We can talk about where this might fit into the lab if you decide to go this route.)

Which option you choose depends on your goal.

One last question that's irrelevant to the rest of my comments: you mention in the PR that your goal is to "help partners be able to integrate their data easier." Does this guide accomplish that goal? I wonder if showing a couple quick examples of sending events is sufficient. What if they want to send metrics, for example? To me, this seems more like a hello world guide on using the Events API.

@mehreentahir16 and I are here to work with you through these decisions, so let me know what you think. If I said anything that doesn't make sense, let me know!

jcountsNR:
Hi @alexronquillo , thanks for the feedback! My goal is to give a how-to guide specifically for partners to set up the data integration with New Relic. This is a bit different than an end user perspective, since things like License key is more specific to each end user. The lab that exists now 'Build a quickstart' effectively skips this step by just having them run an already instrumented docker container.

Option 1 seems like the right way to go actually, and just have it as an optional/alternative intro to the lab. I will have to rewrite it to fit this though. Also, I went with events since that is the most common use case among partners. If you think I should include metrics, logs, and traces, should I include some notes as a sidebar? Or repeat the steps multiple times?

alexronquillo:
> My goal is to give a how-to guide specifically for partners to set up the data integration with New Relic.

@jcountsNR I guess I'm a little confused about the target audience. If a partner is building a quickstart, isn't the assumption that they've already instrumented their software to send data to New Relic. The purpose of quickstarts is to collect that data into a distributable way so end users can quickly see dashboards, alerts, etc. about the data.

If the partner is currently not sending data to New Relic at all, I think that's a big topic with many facets that isn't really covered by a simple how-to guide, IMO. We have many docs on how to get data into New Relic, including those in the "Collect data" directory, such as [this one](https://developer.newrelic.com/collect-data/collect-data-from-any-source/).

jcountsNR:
> isn't the assumption that they've already instrumented their software to send data to New Relic.

Not really, since we are pushing for I/O integration to be a first step in new partnerships. Some of the pre-existing partners will already have this completed, but all new partnerships will need to set this up.



> If the partner is currently not sending data to New Relic at all, I think that's a big topic with many facets that isn't really covered by a simple how-to guide, IMO.

This one you might be right about, and the problem I'm having is thinking of a way to cast a wide enough net on the how to. We are just noticing our biggest blocker to getting partners in our catalog is getting the data integration done, so we wanted to try something different. I could probably do a different guide with just links out to the various ways to ingest data, and maybe that would steer them in the right direction.

alexronquillo:
> I could probably do a different guide with just links out to the various ways to ingest data, and maybe that would steer them in the right direction.

Like [this](https://developer.newrelic.com/collect-data/collect-data-from-any-source/) or something different?

alexronquillo:
> Not really, since we are pushing for I/O integration to be a first step in new partnerships. Some of the pre-existing partners will already have this completed, but all new partnerships will need to set this up.

So, after some reflection, I'm still not sure I follow this. The way I see it, there are two goals:

1. Getting data into New Relic
2. Creating a quickstart

To me, creating a quickstart only comes after getting data into New Relic, which is a task in itself. The existing lab aims to cover goal 2. Namely, assuming you have integrated your app with New Relic, here's how to build a quickstart.

So what's left is goal 1, showing partners how to get data into New Relic. I think this is a big enough topic that it deserves its own content. In my last comment, I linked to a guide that seems to do what you suggested. If it doesn't, let's talk about what exactly you're looking for to help partners accomplish this goal. Are you looking for an example-based tutorial? Are you looking for a narrative-based lab? Are you looking for exhaustive coverage of New Relic telemetry?

Here are some links to check out to move the conversation forward:

- [Data and APIs](https://docs.newrelic.com/docs/data-apis/get-started/nrdb-horsepower-under-hood)
- [Collect data](https://developer.newrelic.com/collect-data/)
- [Event API](https://docs.newrelic.com/docs/data-apis/ingest-apis/introduction-event-api/)
- [Custom data course](https://learn.newrelic.com/custom-data-with-insights)
- [Python SDK (with examples)](https://github.com/newrelic/newrelic-telemetry-sdk-python)

If we can pinpoint what is missing from these resources, it may help us understand what you want to build.

@jcountsNR

jcountsNR:
@alexronquillo , you are spot on with the goal here. The information at bulk seems to be there, but it's been difficult to allow them to self-serve and get data to New Relic in some cases. After this conversation I'm starting to think that a narrative lab might make the most sense, because we already have example tutorials existing, with curl commands and with code examples, and we already have [Collect data](https://developer.newrelic.com/collect-data/) which casts a broad net of examples. 

So maybe what you had mentioned earlier, a flashDB that is not making any POST requests, and we can provide a lab that will set up that data integration from there. Any tips or thoughts for how to get started would be helpful.

jcountsNR:
@alexronquillo , do you want to take a look and see what you think?

src/markdown-pages/contribute-to-quickstarts/instrument-library/index.mdx
```diff
@@ -0,0 +1,20 @@
+---
+title: 'Instrument your Library'
+template: 'OverviewTemplate'
+description: ''
+---
+<Intro>
+If your looking to jump into building your own quickstart, the first step is instrumenting your own library to send data to New Relic. New Relic I/O gives end users the ability to slice and dice observability data around their application, server, etc. Setting up your library to get data into New Relic is critical to this process, so that the end users have a streamlined click-thorugh experience to see your data on our platform. You will need to instrument your library with at least one data type to be able to observe it via a dashboard or alert.
```
alexronquillo, line 7: I would like to see a more thorough explanation of _why_ the developer needs to instrument their library for the purposes of building a quickstart.

> New Relic I/O gives end users the ability to slice and dice observability data around their application, server, etc. Setting up your library to get data into New Relic is critical to this process, so that the end users have a streamlined click-thorugh experience to see your data on our platform.

To be honest, I don't really understand what this means.

> You will need to instrument your library with at least one data type to be able to observe it via a dashboard or alert.

This is more what I'm looking for, but more thorough. I'd like to see this introduction explain things like:

- Why the developer needs to instrument their library on behalf of the user
- How the developer will be able to use their instrumentation to build a quickstart
- etc.

Remember, your target audience here is an onboarding partner, and you want this to be a self-guided course. When you review your piece, it's helpful to ask yourself if you think your target audience understands what you've written enough to move on, without having read much else on the subject. Take your time with explanations. It's much better for you to overexplain than underexplain, and it's much easier to edit as well.

src/markdown-pages/contribute-to-quickstarts/instrument-library/index.mdx
```diff
@@ -0,0 +1,20 @@
...
+</Intro>
+
+## Observability Data Options
+
+ New Relic has [four types of observability data](https://docs.newrelic.com/docs/data-apis/understand-data/new-relic-data-types) which we call MELT data; *Metrics, Events, Logs, and Traces*. Each one handles a specific use case, so you'll have to decide which one fits your needs best.
+
+ Here is some context to help you decide:
+
+ - [Events](contribute-to-quickstarts/create-a-quickstart/instrument-library/send-events) **(Most Common)**: Events are used if you are capturing something occuring in your application, server, library, etc. For CI/CD applications this might be a capture of each time a job runs. For security applications this might be any time a threat is detected.
```
alexronquillo, line 16: Great use of examples here.

src/markdown-pages/contribute-to-quickstarts/instrument-library/index.mdx
```diff
@@ -0,0 +1,20 @@
...
+template: 'OverviewTemplate'
+description: ''
+---
+<Intro>
+If your looking to jump into building your own quickstart, the first step is instrumenting your own library to send data to New Relic. New Relic I/O gives end users the ability to slice and dice observability data around their application, server, etc. Setting up your library to get data into New Relic is critical to this process, so that the end users have a streamlined click-thorugh experience to see your data on our platform. You will need to instrument your library with at least one data type to be able to observe it via a dashboard or alert.
+</Intro>
+
+## Observability Data Options
+
+ New Relic has [four types of observability data](https://docs.newrelic.com/docs/data-apis/understand-data/new-relic-data-types) which we call MELT data; *Metrics, Events, Logs, and Traces*. Each one handles a specific use case, so you'll have to decide which one fits your needs best.
```
alexronquillo, line 12: > Each one handles a specific use case, so you'll have to decide which one fits your needs best.

I'm not sure I follow. How does "metrics" handle a specific use case? Maybe you mean something more like:

> Each one serves a different purpose in providing visibility into your library's behaviors and performance.

Or something like that?

---

> New Relic has [four types of observability data]

I would be careful about jargon like "observability" here. Many people don't know what this means, so a brief definition would be valuable. Later on in this PR, you use others like:

- log drains
- agent
- NRQL
- etc.

These terms may not be clear to your target audience. Sometimes it's best to remove them in favor of something simpler. If you can't, than an explanation is valuable in providing context.

src/markdown-pages/contribute-to-quickstarts/instrument-library/index.mdx
```diff
@@ -0,0 +1,20 @@
...
+
+ Here is some context to help you decide:
+
+ - [Events](contribute-to-quickstarts/create-a-quickstart/instrument-library/send-events) **(Most Common)**: Events are used if you are capturing something occuring in your application, server, library, etc. For CI/CD applications this might be a capture of each time a job runs. For security applications this might be any time a threat is detected.
+ - [Logs](https://docs.newrelic.com/docs/logs/log-api/introduction-log-api/) **(Common)**: Logs should be used if your library uses multiple attributes (ie. nested key-value data), or if your library is set up already to use log drains.
+ - [Metrics](https://docs.newrelic.com/docs/data-apis/ingest-apis/metric-api/introduction-metric-api/) **(Uncommon)**: Metrics are generally collected as aggregated data or as a numeric status at a point in time. Most metrics can be sent or aggregated through events, which is why we recommend events instead. You might want to use metrics if you view the data at a very high level, and don't expect your users to need a more detailed analysis.
```
alexronquillo, line 18: Do you need to specify which of these are common and uncommon? Can we just let the partner decide which ones would be valuable for them?

src/markdown-pages/contribute-to-quickstarts/instrument-library/send-events.mdx
```diff
@@ -0,0 +1,170 @@
...
+title: 'Send Events from Your Library'
+template: 'GuideTemplate'
+description: 'With our Events API you can ingest your data as the first step towards your quickstart contribution.'
+tags:
+  - quickstart
+---
+
+<Intro>
+
+This guide will walk you through how to instrument your application to send event data into the New Relic's [Event API](https://docs.newrelic.com/docs/data-apis/ingest-apis/introduction-event-api/). The Event API allow you to send events to New Relic without a need for an agent. First, let's cover what events are. An event is a record of something happening somewhere. What triggers the event and what data is captured is up to you to customize, which we'll discuss next.
```
alexronquillo, line 11: > An event is a record of something happening somewhere.

This is pretty vague. Could you provide an example?

src/markdown-pages/contribute-to-quickstarts/instrument-library/send-events.mdx
```diff
@@ -0,0 +1,170 @@
...
+$gzipStream.Write($enc_data, 0, $enc_data.Length)
+$gzipStream.Close()
+$gzipBody = $output.ToArray()
+
+Invoke-WebRequest -Headers $headers -Method Post -Body $gzipBody  "https://insights-collector.newrelic.com/v1/accounts/events"
+```
+
+
+
+### Using our SDK's
```
alexronquillo, line 72: I'm a little confused why you changed these headings from the outline?

src/markdown-pages/contribute-to-quickstarts/instrument-library/send-events.mdx
```diff
@@ -0,0 +1,170 @@
...
+
+<Intro>
+
+This guide will walk you through how to instrument your application to send event data into the New Relic's [Event API](https://docs.newrelic.com/docs/data-apis/ingest-apis/introduction-event-api/). The Event API allow you to send events to New Relic without a need for an agent. First, let's cover what events are. An event is a record of something happening somewhere. What triggers the event and what data is captured is up to you to customize, which we'll discuss next. 
+
+</Intro>
+
+## Capture Events in your Library
+
+At New Relic, we report events to data objects also called events. These events have multiple attributes (key-value pairs). We name each event record in our system with the event-type key.
```
alexronquillo, line 17: > At New Relic, we report events to data objects also called events.

I don't understand what this means. Could you provide some more useful explanation around this concept?

> We name each event record in our system with the event-type key.

In what way is that a "name"? Do you mean to say:

> Each event record you send must have an `event-type` attribute.

or something else?

src/markdown-pages/contribute-to-quickstarts/instrument-library/send-events.mdx
```diff
@@ -0,0 +1,170 @@
...
+
+</Intro>
+
+## Capture Events in your Library
+
+At New Relic, we report events to data objects also called events. These events have multiple attributes (key-value pairs). We name each event record in our system with the event-type key. 
+```json
+[{"event-type": "transaction"}]
+```
+The event-type key is how you will query the event data from NRQL. using the above example, this is how we would query for the data in NRQL:
```
alexronquillo, line 21: I'm confused why you are you talking about NRQL here, and your readers probably will be as well. Could you explain this a bit more? This document is about getting data into New Relic. I think you can probably leave off most low-level mentions of accessing that data until the part where you build a quickstart. What do you think?

src/markdown-pages/contribute-to-quickstarts/instrument-library/send-events.mdx
```diff
@@ -0,0 +1,170 @@
...
+  - quickstart
+---
+
+<Intro>
+
+This guide will walk you through how to instrument your application to send event data into the New Relic's [Event API](https://docs.newrelic.com/docs/data-apis/ingest-apis/introduction-event-api/). The Event API allow you to send events to New Relic without a need for an agent. First, let's cover what events are. An event is a record of something happening somewhere. What triggers the event and what data is captured is up to you to customize, which we'll discuss next. 
+
+</Intro>
+
+## Capture Events in your Library
```
alexronquillo, line 15: I'm confused why you have this section. It talks about a few things:

- That events have attributes (this is important)
- That events have an event-type attribute (this doesn't seem important, considering none of your examples later on set or make use of this attribute)
- That events can be queried on their event-type attribute (this really doesn't seem important in this document)

It seems like you could bring the first point up into the introduction and drop the rest of this section to me. Thoughts?

src/markdown-pages/contribute-to-quickstarts/instrument-library/send-events.mdx
```diff
@@ -0,0 +1,170 @@
...
+The event-type key is how you will query the event data from NRQL. using the above example, this is how we would query for the data in NRQL:
+
+```sql
+From transaction select *
+```
+The rest in the event can be any custom key-value data, based on your library. When preparing the payload, consider the use case that makes the most sense for your users, and build the events accordingly. Here are some more resources about event data if you need more information.
+-  [Events at New Relic](https://docs.newrelic.com/docs/data-apis/understand-data/new-relic-data-types/#events-new-relic)
+-  [Formatting JSON](https://docs.newrelic.com/docs/data-apis/ingest-apis/introduction-event-api#instrument)
+
+## Send your events to New Relic
```
alexronquillo, line 30: This is what the whole document is really about, so this should be up in the introduction.

src/markdown-pages/contribute-to-quickstarts/instrument-library/send-events.mdx
```diff
@@ -0,0 +1,170 @@
...
+There are a few ways we will discuss in order to send your event data into New Relic. We have our [telemetry SDK's](https://docs.newrelic.com/docs/data-apis/ingest-apis/telemetry-sdks-report-custom-telemetry-data), which are development kits instrumented to help you easily set up a POST request via several very common languages (C, Java, Python, .NET, etc.). We also have a basic curl command, which can be used on a windows or linux system to make a POST request to New Relic. We also have a prometheus option, if you use prometheus and have difficulty with the first two options.
```
alexronquillo, line 31: I have a few comments about this paragraph:

- Can you order these methods the same order that you use for the sections?
- Can you define SDK before using the acronym?
- Can you limit the languages you refer to to the ones you provide examples for?
- Can you make this into a list instead of a paragraph?

src/markdown-pages/contribute-to-quickstarts/instrument-library/send-events.mdx
```diff
@@ -0,0 +1,170 @@
...
+
+### New Relic License Key
+
+You will need to set up an environment variable for your users to input their [New Relic License key](https://docs.newrelic.com/docs/apis/intro-apis/new-relic-api-keys/#ingest-license-key), or give an option for them to add it to the request header. The examples below will assume this variable is configured as `NEW_RELIC_INSERT_KEY`.
```
alexronquillo, line 35: Can you show them how to do this? If it's different for each method, talk about it in the sections below instead of in a section of its own.

src/markdown-pages/contribute-to-quickstarts/instrument-library/send-events.mdx
```diff
@@ -0,0 +1,170 @@
...
+
+### Send events via a curl/PowerShell command
+
+First, let's cover how you can send agentless data to New Relic, with just a simple command. These commands can be used in Linux or Windows environments. Here are some example requests for both
```
alexronquillo, line 39: I don't see much value in just showing how to format a curl request. Can you cover why this is a useful method for partners? Can you provide a more fleshed-out example of how they might use this approach? Try to lean on your experienced partners needing in the past to help future partners.

src/markdown-pages/contribute-to-quickstarts/instrument-library/send-events.mdx
```diff
@@ -0,0 +1,170 @@
...
+
+New Relic has a set of open source API client libraries that send data to New Relic. These operate by using our API's for data ingest, and in this case our event API. We have two SDK's that work with events, and those are Python and Java. Here we will walk through how to install each, and give a code example using both SDK's.
+
+For Python, you'll need to install the `newrelic-telemetry-sdk` package. you can do this by using pip:
+
+```bash
+$ pip install newrelic-telemetry-sdk
+```
+For Java, you will need to set up several dependencies:
+```java
```
alexronquillo, line 82: Is there anything else the user has to do to install the dependencies? Can you provide steps for that? Also, have you tried out all this code to make sure it works?

src/markdown-pages/contribute-to-quickstarts/instrument-library/send-events.mdx
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  private static void sendSampleEvent(
+      TelemetryClient telemetryClient, Attributes commonAttributes) {
+    Event event = new Event("TestEvent", new Attributes().put("testKey", "testValue"));
+    telemetryClient.sendBatch(new EventBatch(singleton(event), commonAttributes));
+  }
+
+```
+### Alternative Options
```
alexronquillo, line 165: One option you haven't covered is just making API requests with the developer's favorite http library (in other words, manual instrumentation). Is this not something you see partners doing?

alexronquillo (commented):
Thanks for the new commit @jcountsNR Please see my comments below. Let me know if you have any pushback.

src/markdown-pages/contribute-to-quickstarts/instrument-library/index.mdx
jcountsNR: I don't have to, I thought it might be a way to point them towards events or logs which is really what makes the most sense. I think if the descriptions are thorough enough that should be clear though.

src/markdown-pages/contribute-to-quickstarts/instrument-library/send-events.mdx
jcountsNR: That's true, I guess I should just direct them to our event API page in that case?

src/markdown-pages/contribute-to-quickstarts/instrument-library/send-events.mdx
jcountsNR: Just updating on this. They don't need to do anything other than installing and running gradle build, so I actually merged the decencies with the code block and just mentioned running gradle. I have run the Python code and it works, I am still trying to set up java and the dependencies to run the java snip but it's a direct example from our [example events docs](https://github.com/newrelic/newrelic-telemetry-sdk-java/blob/main/telemetry_examples/src/main/java/com/newrelic/telemetry/examples/EventExample.java) so it should work.

jcountsNR:
@alexronquillo , PR commit added. Let me know what you think.

src/markdown-pages/contribute-to-quickstarts/instrument-library/send-events.mdx
jcountsNR: I added some better context here, so I'll resolve this.

src/markdown-pages/contribute-to-quickstarts/instrument-library/send-events.mdx
jcountsNR: I added something for this.

src/markdown-pages/contribute-to-quickstarts/instrument-library/send-events.mdx
jcountsNR: Done.

src/markdown-pages/contribute-to-quickstarts/instrument-library/send-events.mdx
jcountsNR: This was updated with the recent commits, but I'll leave this open. until it's reviewed.

src/markdown-pages/contribute-to-quickstarts/instrument-library/index.mdx
jcountsNR: I believe I was able to resolve these issues with the last commit, but I'll leave this open.

src/markdown-pages/contribute-to-quickstarts/instrument-library/index.mdx
jcountsNR: I overhauled this paragraph and added some more explanation.

src/markdown-pages/contribute-to-quickstarts/instrument-library/send-events.mdx
```diff
@@ -0,0 +1,192 @@
...
+```
+Next you can use the event client to send your event data
+```Python
+import os
+import time
+from newrelic_telemetry_sdk import Event, EventClient
+
+#Replace with your event data
+event = Event(
+    "RateLimit", {"path": "/v1/endpoint", "accountId": 1000, "rejectRatio": 0.1}
```
alexronquillo, line 108: `accountId` should also be a variable, no?

alexronquillo: Oh, nevermind, I think I understand. This is part of the payload.

alexronquillo (approved):
@jcountsNR Thanks for submitting this guide! I've reviewed and edited your work and will merge it into develop. Please feel free to take a look at the edits!

We'll take care of reorganizing the existing pieces to work all the quickstarts docs into a more uniform course.

PR #2024, opened 2022-01-18
Typescript Configuration for developer website

josephgregoryii:
## Dev Dependencies
- Node Types
- React Types
- React-DOM types
- [gatsby-plugin-typegen](https://www.gatsbyjs.com/plugins/gatsby-plugin-typegen/)
- typescript
- ts-loader
- source-map-loader

- typescript-eslint/parser
- typescript-eslint/eslint-plugin

### Notes:
- When linting via command line, there will be a *ton* of linter warnings. This will slowly decrease as we adapt typescript formatting.
- Linter should work as previously intended. 
- We may also turn `on` some linter options at a later point that were needed to be turned `off` to allow the same rules as previous.

### Suggestions:
- I would suggest sandboxing in the typescript environment and see how the linter feels with your editor.


clarkmcadoo (dismissed):

package.json
```diff
@@ -26,6 +28,7 @@
     "gatsby-plugin-sass": "^4.8.0",
     "gatsby-plugin-sharp": "^3.14.1",
     "gatsby-plugin-sitemap": "^4.4.0",
+    "gatsby-plugin-typegen": "^2.2.4",
```
rudouglas, line 31: Should these be moved to `devDependencies`?

rudouglas (changes requested):
Just a comment on moving packages to `devDependencies` otherwise this looks awesome

package.json
```diff
@@ -54,6 +57,8 @@
     "react-use": "^15.3.4",
     "sharp": "^0.29.1",
     "snyk": "^1.437.4",
+    "source-map-loader": "^3.0.1",
+    "ts-loader": "^9.2.6",
```
aswanson-nr, line 61: These should be dev dependencies in a normal react app, please disregard if that's not the case for a Gatsby site.

aswanson-nr (commented):
Looks great!! Just saw some deps that should probably be dev dependecies

package.json
josephgregoryii: done :)

aswanson-nr (approved):

PR #2033, opened 2022-01-26
feat: new styles for subtask in lab

kisielewskik:
## Description

Task from Alex Ronquillo from New Relic org. Purpose of this is making new styles for lab procedures while we have parrarel sub  procedures.

## Reviewer Notes

For testing it you have to make procedure with not-integer ProdIdx, example: ProcIdx 2.1, should be under ProcIdx 2 with simpler styles.

### DIFF
--- a/src/components/GuideTile/GuideTile.js
+++ b/src/components/GuideTile/GuideTile.js
@@ -22,6 +22,7 @@ const GuideTile = ({ duration, title, description, className, to }) => (
     <h3
       css={css`
         font-size: 1rem;
+        margin-bottom: 0.5rem;
       `}
     >
       {title}
--- a/src/components/GuideTile/GuideTile.js
+++ b/src/components/GuideTile/GuideTile.js
@@ -9,7 +9,7 @@ const GuideTile = ({
   description,
   className,
   to,
-  procIdxisNotInteger,
+  procIdxIsInteger,
 }) => (
   <Surface
     as={Link}
@@ -20,8 +20,8 @@ const GuideTile = ({
     css={css`
       display: grid;
       grid-template-rows: auto 1fr auto;
-      width: ${procIdxisNotInteger ? '100%' : '90%'};
-      margin-left: ${procIdxisNotInteger ? '0' : '10%'};
+      width: ${procIdxIsInteger ? '100%' : '90%'};
+      margin-left: ${procIdxIsInteger ? '0' : '10%'};
       border-radius: 0.25rem;
       position: relative;
       padding: 1rem;
@@ -31,7 +31,7 @@ const GuideTile = ({
     <h3
       css={css`
         font-size: 1rem;
-        margin-bottom: ${procIdxisNotInteger ? '1rem' : '0.1rem'};
+        margin-bottom: ${procIdxIsInteger ? '1rem' : '0.1rem'};
       `}
     >
       {title}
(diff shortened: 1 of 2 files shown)


DominikMarciniszyn (dismissed):

src/templates/LabOverviewTemplate.js
```diff
@@ -47,21 +47,29 @@ const LabOverviewTemplate = ({ data, location }) => {
...
-                      description={
-                        frontmatter.tileShorthand?.description ||
-                        frontmatter.description
-                      }
-                    />
-                  ))}
+                  .map(({ fields, frontmatter }, index) => {
+                    const procIdxisNotInteger = Number.isInteger(
+                      frontmatter.procIdx
+                    );
```
LizBaker, line 53: as this is set up now procIdxisNotInteger will evaluate to true when a number is an integer. would you mind renaming it to procIdxIsInteger?

kisielewskik: I changed name of variable is more logic now

src/components/GuideTile/GuideTile.js
```diff
@@ -13,6 +20,8 @@ const GuideTile = ({ duration, title, description, className, to }) => (
     css={css`
       display: grid;
       grid-template-rows: auto 1fr auto;
+      width: ${procIdxIsInteger ? '100%' : '90%'};
+      margin-left: ${procIdxIsInteger ? '0' : '10%'};
```
LizBaker, line 24: changing the styles here have had some unintended consequences in other places that we use the GuideTiles. I think moving these styles to the LabOverviewTemplate might be best. you can use the css prop within that guidetile just like here and select for the p and h3

normal styling:
Screen Shot 2022-02-04 at 9 15 26 AM

current changes:
Screen Shot 2022-02-04 at 9 15 34 AM

jpvajda: That spacing between the tiles seems really wide I'd recommend following the pattern we use for space between the quickstart tiles on I/O 

https://developer.newrelic.com/instant-observability/
Screen Shot 2022-02-04 at 11 00 32 AM

jpvajda (changes requested):

src/components/GuideTile/GuideTile.js
kisielewskik: Totalny agree, I didn’t noticed that change will have so big impact on other views, i will move that styles to LabOverviewTemplate asap to keep website pattern, thank you for your advices

src/templates/LabOverviewTemplate.js
```diff
@@ -47,21 +47,29 @@ const LabOverviewTemplate = ({ data, location }) => {
...
-                      title={`${frontmatter.procIdx}. ${
-                        frontmatter.tileShorthand?.title || frontmatter.title
-                      }`}
-                      description={
-                        frontmatter.tileShorthand?.description ||
-                        frontmatter.description
-                      }
-                    />
-                  ))}
+                  .map(({ fields, frontmatter }, index) => {
```
jpvajda, line 50: Attached is the screen shot for this change so it's visible, this is OK for me, so I approve this.
Screen Shot 2022-02-04 at 9 47 45 AM

jpvajda (changes requested):

src/components/GuideTile/GuideTile.js
jpvajda: Thanks @kisielewskik!

LizBaker (approved):

PR #2045, opened 2022-02-08
fix: update IO url across site

clarkmcadoo:


zstix (approved):
I'm not sure if we want to have `related-pages.json` in `develop`, but that's outside the scope of this PR.

zstix (changes requested):
Let's remove the changes to `related-pages.json`

zstix (approved):

PR #2066, opened 2022-02-22
update lock file

nr-kkenney:
update lock file since it got out of sync with previous change to package.json


nr-kkenney:
closing in favor of #2067

PR #2072, opened 2022-03-04
fix: add 301 statusCode to IO site redirect

clarkmcadoo:


gatsby-node.js
```diff
@@ -37,6 +37,7 @@ const createRedirectsForNewIoDomain = (createRedirect, allQuickstarts) => {
     toPath: 'https://www.newrelic.com/instant-observability',
     redirectInBrowser: true,
```
zstix, line 38: We're not sure, but this _might_ help with the issue:

```suggestion
```

zstix:
@clarkmcadoo Can you provide a little more context on why this change was made. Thanks!

clarkmcadoo:
Got a request from an SEO expert wanting to update the 301 redirect from developer.newrelic.com/instant-observability to the new IO site from a meta refresh redirect to a true 301 redirect.

Slack -> https://newrelic.slack.com/archives/C01R2TB8BUZ/p1646340632880059

aswanson-nr (approved):

PR #2078, opened 2022-03-07
fix: remove gatsby-plugin-meta-redirect from project

clarkmcadoo:
`gatsby-plugin-meta-redirect` was making all of the redirects via `html` instead of a server redirect


nr-kkenney (dismissed):

aswanson-nr:
We'll want to confirm if all the other redirects on this site continue to work properly. Especially around embedded guides

PR #2079, opened 2022-03-08
Clark/make io 301 redirects

clarkmcadoo:
This PR solves the splat redirects as well as our issue with having true 301 redirects


clarkmcadoo:
This is testable at: https://developerwebsiteclarkmakeio301.gtsb.io/instant-observability

Screen Shot 2022-03-09 at 10 46 03 AM

aswanson-nr:
Since this adds the splat redirects, could you also remove [this](https://github.com/newrelic/developer-website/blob/develop/gatsby-node.js#L101-L109) part of the graphql query and the `src/data/quickstarts.json` file?

clarkmcadoo:
@aswanson-nr I was able to do most of your requests, but some of the plugins in the gatsby config still seem to be needed, the site wouldn't build without them.. Probably could audit this in the future and remove all the unnecessary ones in a different ticket one day?

gatsby-config.js
```diff
@@ -167,13 +167,6 @@ module.exports = {
...
     },
-    {
-      resolve: 'gatsby-source-filesystem',
-      options: {
-        name: 'quickstarts',
-        path: `${__dirname}/src/data/quickstarts.json`,
-      },
-    },
     {
       resolve: 'gatsby-transformer-json',
```
aswanson-nr, line 171: One last thing to remove! This transformer is only used for the quickstart json objects

aswanson-nr (dismissed):

zstix (approved):

zstix (approved):

PR #2174, opened 2022-09-23
Remove 'new relic one'

zuluecho9:
Removing instances of 'New Relic One application', 'New Relic One app' and replacing with 'New Relic app', and replacing references to 'New Relic One' platform context as just 'New Relic'.

What I didn't change:
New Relic One CLI (as this is still in codebase, but we explain that in doc that it is about building New Relic apps)
New Relic One SDK (as that's also in codebase still)

It was thought these references were less important and could stay as-is for now, based on conversations.


jaesius (commented):
Only one small edit is needed

PR #2189, opened 2022-11-13
chore: update to changelog playlist

mehreentahir16:


jaesius (changes requested):
The embedded Youtube video is pushed outside the right margin causing a horizontal scroll on the page.

jaesius (approved):

PR #2205, opened 2023-01-30
updating front page for docs

akristen:
- removed yarn 3.2.0 pkg manager from file
- Added css magic to some file at some point
- Changed out I/O video for generic observability video
- Updated content from I/O to links to the docs site

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -13,7 +13,6 @@ import FeaturedGuideTile from '../components/FeaturedGuideTile';
 import { PageContext } from '../components/PageContext';
 import { pageContext } from '../types';
 import * as styles from './index.module.scss';
-import Video from '../components/Video';
 import devChampionBadge from '../images/developer-champion/dev-champion-badge.png';
 import podcastBadge from '../images/podcasts/podcasts-badge.png';
 const getStartedGuides = [
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -66,10 +66,33 @@ const IndexPage = ({ data, pageContext, location }) => {
           >
             <div className={styles.introText}>
               <p>
-                New Relic lets you bring in data from any digital source so that you can fully understand your system, analyze that data efficiently, and respond to incidents before they become problems. 
+
+New Relic is an observability platform that helps you accurately capture data critical to the success of your team. Whether you're a developer installing your first agent or you're ready to query ingested data, our docs site can help you find the best solution for your performance needs:
+<p>
+</p>
+<p className={styles.inspiration}>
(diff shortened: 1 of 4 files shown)


src/pages/index.js
```diff
@@ -66,53 +61,43 @@ const IndexPage = ({ data, pageContext, location }) => {
...
-                <br />
-                <br />
-                Didn't find what you're looking for? You're always{' '}
-                <a href="https://github.com/newrelic/newrelic-quickstarts">
-                  welcome to add contributions
-                </a>{' '}
-                or improve existing quickstarts to help others instantly monitor
-                their services.
+                New Relic is an observability platform that helps you accurately
+                capture data critical to the success of your team. Whether
```
bradleycamacho, line 65:
```suggestion
                New Relic is an observability platform that helps you
                capture data critical to the success of your team. Whether
```

bradleycamacho: Not sure this first sentence is the best sell of the platform.

src/pages/index.js
```diff
@@ -66,53 +61,43 @@ const IndexPage = ({ data, pageContext, location }) => {
...
+                you're a developer installing your first agent or you're ready
+                to query ingested data, our docs site can help you find the best
+                solution for your performance needs:
```
bradleycamacho, line 68: Appealing to different audience is great here, but I think there's a bit of whiplash going on.

The user lands on the dev site, but then instantly gets talked to about the docs site. That's the goal, but I think this intro should try to orient them a bit about why we're linking them over to the docs site AND specifically mention the docs site is a different site.

bradleycamacho (changes requested):

bradleycamacho:
Screen Shot 2023-01-30 at 3 35 16 PM

I'm also a bit iffy on the general look of the homepage with this content. I wonder if we should add an image/video to the right side of the content. Maybe a big signup button? Just seems a bit empty and not whimsical.

src/pages/index.js
akristen: Yeah, it's a bit of a hodge podge from existing content (blog post about DevOps/New Relic + Welcome to NR doc). Wanted it to be specific to what a DevOps team might need and guide them to the docs site for their answers

src/pages/index.js
akristen: maybe changing 'our' in the links to 'Quickstart' blahblahblah to 'the Quickstart blahblahblah on the New Relic docs site' instead. Or some kind of 'docs site' call out specifically.

bradleycamacho (approved):

PR #2206, opened 2023-01-31
feat(sonic): port terraform and add redirects

bradleycamacho:
This PR removes the terraform docs that are now ported to the docs site and adds redirects to the docs site.


akristen (approved):
Approved! Checked the redirect and it worked!

PR #2208, opened 2023-02-13
fix(sonic): fix broken terraform doc external redirects

bradleycamacho:
Fixes external redirects introduced in #2206

### DIFF
--- a/src/data/external-redirects.json
+++ b/src/data/external-redirects.json
@@ -24,16 +24,20 @@
     "paths": ["/collect-data/monitor-your-application/estimate-cost"]
   },
   {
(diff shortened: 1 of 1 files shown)


roadlittledawn (commented):
Why redirecting to docs URLs with no trailing slash?

bradleycamacho:
> Why redirecting to docs URLs with no trailing slash?

I just noticed that's what we do for the rest of the redirects so I just followed the same pattern!

jeff-colucci (approved):
Looks good to me

akristen (approved):
Lin ks work!

PR #2286, opened 2024-06-12
Add redirects and remove pages

LizBaker:


paperclypse (approved):
It built, the redirects work, looking good! Thanks for all your help @LizBaker!

PR #2306, opened 2026-07-30
fix(security): use URL hostname check for isNewRelicDomain + exclude vendor minified files from CodeQL

pranav-new-relic:
## What and Why

This PR fixes three CodeQL `js/incomplete-url-substring-sanitization` alerts in [NR-560205](https://new-relic.atlassian.net/browse/NR-560205).

**Background (for those new to this):** When you check whether a URL belongs to a certain website by looking for substrings like `url.includes("newrelic.com/")`, an attacker can craft a URL like `https://evil.com/newrelic.com/docs` that passes the check even though it clearly belongs to `evil.com`. The safest way to check a URL's true domain is to actually *parse* it using the browser's built-in `URL` class.

## Changes

**1. `src/@newrelic/gatsby-theme-newrelic/components/Link.js`** — fixes alerts [#1](https://github.com/newrelic/developer-website/security/code-scanning/1) and [#2](https://github.com/newrelic/developer-website/security/code-scanning/2)

The `isNewRelicDomain` helper previously used:
- `to.endsWith("newrelic.com")` — matched `https://evilnewrelic.com` ❌
- `to.includes("newrelic.com/")` — matched `https://evil.com/newrelic.com/page` ❌

It now parses the URL with `new URL(to)` and checks the actual **hostname** field, so those crafted URLs are correctly rejected. A `try/catch` handles non-URL inputs gracefully.

**2. `.github/codeql/codeql-config.yml`** *(new file)* — closes alert [#3](https://github.com/newrelic/developer-website/security/code-scanning/3)

Alert #3 fires in `static/tessen.min-1.14.0.js`, a pre-built vendored analytics library. Editing minified files directly is impractical; this config tells CodeQL to skip `static/**/*.min*.js` files, which are build artifacts rather than first-party source.

## Testing

`isNewRelicDomain` is used on line 102 to decide whether to include `noreferrer` in the `rel` attribute for external links. The fix is a strict improvement: real New Relic domains still work, and crafted URLs are correctly blocked. The try/catch ensures no breakage on unusual inputs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[NR-560205]: https://new-relic.atlassian.net/browse/NR-560205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/@newrelic/gatsby-theme-newrelic/components/Link.js
```diff
@@ -14,8 +14,14 @@ import { css } from '@emotion/react';
 
 const isHash = (to) => to.startsWith('#');
 const isExternal = (to) => to.startsWith('http');
-const isNewRelicDomain = (to) =>
-  to.endsWith('newrelic.com') || to.includes('newrelic.com/');
+const isNewRelicDomain = (to) => {
+  try {
+    const { hostname } = new URL(to);
+    return hostname === 'newrelic.com' || hostname.endsWith('.newrelic.com');
```
pranav-new-relic, line 20: **Why this check works when the old one didn't**

The original code used substring checks:
- `to.endsWith('newrelic.com')` → passes for `https://evilnewrelic.com` ❌  
- `to.includes('newrelic.com/')` → passes for `https://evil.com/newrelic.com/path` ❌

A URL has distinct parts: *scheme* (`https://`), *hostname* (`evil.com`), and *path* (`/newrelic.com/path`). By parsing the URL with the built-in `URL` constructor we get the **hostname** field — the actual server the browser connects to — which an attacker cannot forge while still reaching `evil.com`:

```
new URL('https://evilnewrelic.com').hostname   → 'evilnewrelic.com'  → no match ✅
new URL('https://evil.com/newrelic.com/p').hostname → 'evil.com'     → no match ✅
new URL('https://docs.newrelic.com/guide').hostname → 'docs.newrelic.com' → .endsWith('.newrelic.com') ✅
```

The `try/catch` is there because `new URL(x)` throws a TypeError if `x` is not a valid URL (e.g. a bare path like `/docs`). This function is only called for external links (those starting with `http`) so in practice it will never throw — but the guard keeps things safe for future callers.

akashreddy-nr: LGTM

PR #2312, opened 2026-08-13
fix: redirect EOL developer-website traffic to docs.newrelic.com

akashreddy-nr:
## Summary
- The site has been past its EOL date, and production currently serves a static "This site has been deprecated" splash page instead of an actual redirect to docs.newrelic.com.
- Adds a forced catch-all redirect (`/* → https://docs.newrelic.com/*`), registered **after** all existing curated per-page redirects in `gatsby-node.js` so those keep taking priority — verified this doesn't shadow existing mappings (e.g. `/automate-workflows/get-started-new-relic-cli/` still lands on its specific docs page).
- `force: true` is required so the redirect overrides the existing built `index.html` (which is what's currently serving the splash content) instead of Netlify preferring that static file.

## Test plan
- [ ] Confirm `gatsby build` generates the expected `_redirects` entries, with the catch-all listed last
- [ ] Deploy preview: verify `/` and an arbitrary unmapped path 301 to `docs.newrelic.com`
- [ ] Deploy preview: verify a known curated redirect (e.g. `/automate-workflows/get-started-new-relic-cli/`) still resolves to its specific docs.newrelic.com target, not the catch-all

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Nandu-pns:
Checked the ordering/splat mechanics against Netlify + Gatsby source — no correctness issues:
- **Ordering is safe**: Netlify's `_redirects` is first-match-wins, and the site's own `gatsby-node.js` (`default-site-plugin`) always loads *last* among plugins, so `gatsby-plugin-embed-pages`'s redirects land earlier in the file and won't be shadowed.
- **Splat syntax is correct**: `gatsby-plugin-netlify` converts trailing `*` in `toPath` to `:splat`, so this compiles to a valid rule — same pattern as the existing `/instant-observability/*` redirect.

One thing worth a deliberate call: since the splat is preserved, unmapped paths redirect to the *same path* on docs.newrelic.com. I spot-checked a couple (`/components/table`, `/build-apps/build-hello-world-app/`) directly and both 404 there — developer-website's URL structure doesn't map onto docs.newrelic.com's, which is why the curated list above this exists. So most non-curated traffic will land on a docs 404/search page rather than the docs homepage. Still better than the current splash page, so not blocking — but might be worth simplifying to a flat `toPath: 'https://docs.newrelic.com'` unless there's reason to expect paths resolve 1:1.

Also: `force: true` on `/*` catches `robots.txt`/`sitemap.xml`/static assets too — probably fine for an EOL'd site, just flagging.

